PR #6, opened 2015-03-14
Serve Static Assets

hone:
Allow the fastboot server to serve static assets so you don't need to run a separate ember server to get your fastboot app working like in the emberconf keynote. /cc @rwjblue

### DIFF
--- a/lib/commands/fastboot.js
+++ b/lib/commands/fastboot.js
@@ -21,6 +21,7 @@ module.exports = {
     });
 
     var app = express();
+    app.get('/', server.middleware());
     app.use(express.static('dist'));
     app.get('/*', server.middleware());
 
--- a/lib/commands/fastboot.js
+++ b/lib/commands/fastboot.js
@@ -21,6 +21,7 @@ module.exports = {
     });
 
     var app = express();
+    app.use(express.static('dist'));
     app.get('/*', server.middleware());
 
     var ui = this.ui;
--- a/lib/models/server.js
+++ b/lib/models/server.js
@@ -1,4 +1,3 @@
-var express = require('express');
 var chalk = require('chalk');
 var fs = require('fs');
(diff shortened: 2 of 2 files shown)


hone:
This is still a WIP.

rwjblue:
Lookin' good!  Will take it for a spin in a bit....

hone:
I've run into a few issues, but got a demo of it working: <https://ember-fastboot-demo.herokuapp.com/>

tomdale:
Nice job.

stefanpenner:
> I've run into a few issues, but got a demo of it working: https://ember-fastboot-demo.herokuapp.com/. Source code

maybe I am pointing out the obvious, but I literally see 2 versions of the todo app.

screen shot 2015-03-15 at 12 11 14 pm

patricksimpson:
@stefanpenner because the static files are being served, and then ember runs in the client which produces the site again. A hack I am working on was to delete the existing static HTML before Ember rebuilds it.

stefanpenner:
> @stefanpenner because Ember runs in the client which produces the site again. A hack I am currently using was to delete the existing static HTML before Ember renders it.

yes, but the final goal is for the ember app to "consume" this pre-prendered section, and not-render again.

tomdale:
The full solution is gated on getting Glimmer merged in. The hack is actually surprisingly usable, but I would hate to see what happens when it takes more than a few hundred milliseconds for the JS to finish loading.

stefanpenner:
> The full solution is gated on getting Glimmer merged in. The hack is actually surprisingly usable, but I would hate to see what happens when it takes more than a few hundred milliseconds for the JS to finish loading.

thanks for that tidbit, I assumed as much. Keep up the good work!

rwjblue:
These commits got pulled into #20, and have now landed...

PR #7, opened 2015-03-15
Update to ember-cli 0.2.0

ef4:
Updating to the latest ember-cli release. Motivated by upstream breakage in glob and/or rimraf that broke the build when I tried to install from scratch.


rwjblue:
:+1:

PR #8, opened 2015-03-15
[WIP] Update own dummy app for fastboot compatibility

ef4:
Running `ember fastboot` inside the ember-cli-fastboot repo should boot the dummy app in fastboot mode, analogous to the way `ember serve` or `ember build` work in an addon. This would make it much easier for us to build out a test suite.

There are two remaining blockers for this behavior right now: 

 - `findAppFile` fails to discover `dummy.js` because it's looking for `ember-cli-fastboot.js` instead
 - `findHTMLFile` looks in hardcoded paths which are not correct for the dummy app.

If you hack around both of the above, it boots!

I'm not sure if ember-cli exports appropriate APIs to solve the above issues cleanly. We can certainly hack around them, but I want to get some feedback first.

### DIFF
--- a/Brocfile.js
+++ b/Brocfile.js
@@ -2,7 +2,9 @@
 
 var EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
-var app = new EmberAddon();
+var app = new EmberAddon({
+  storeConfigInMeta: false
+});
 
 // Use `app.import` to add additional libraries to the generated
 // output files.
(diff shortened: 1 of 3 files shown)


ef4:
@rwjblue is there an API that a command like `ember fastboot` could use to detect that it's being run inside an addon and should therefore run against the dummy app instead? 

In particular, it should see the app name as "dummy" and a project root path of `the-addon/tests/dummy` instead of `the-addon`.

rwjblue:
@ef4 `this.project.isEmberCLIAddon()` vs `this.project.isEmberCLIProject()`

ef4:
Great, that should do it, thanks.

PR #13, opened 2015-04-05
Enable content-for from other addons

gdub22:
Any other addons that use the contentFor hook to inject content into the html file are not respected by fastboot.  This change makes fastboot read the **built** index html.  Not sure if this is the best approach. Open to other ideas.

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,5 +8,13 @@ module.exports = {
     return {
       fastboot: require('./lib/commands/fastboot')
     };
+  },
+
+  contentFor: function(type, config) {
+    // When using a dist built html file, we need to reinsert 
+    // content-for body so fastboot can insert its content.
+    if (type === 'body') {
+      return "{{content-for 'body'}}";
+    }
   }
 };
--- a/lib/commands/fastboot.js
+++ b/lib/commands/fastboot.js
@@ -75,11 +75,9 @@ function findHTMLFile() {
   var fs = require('fs');
   if (fs.existsSync('app/fastboot.html')) {
     return 'app/fastboot.html';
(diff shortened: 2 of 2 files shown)


index.js
```diff
@@ -8,5 +8,13 @@ module.exports = {
     return {
       fastboot: require('./lib/commands/fastboot')
     };
+  },
+
+  contentFor: function(type, config) {
+    // When using a dist built html file, we need to reinsert 
+    // content-for body so fastboot can insert its content.
+    if (type === 'body') {
+      return "{{content-for 'body'}}";
```
rwjblue: Can you tweak this and the [associated replacement function](https://github.com/tildeio/ember-cli-fastboot/blob/038a71dc3d64901bfb14e5d637a57651f69ce895/lib/models/server.js#L24-L31) to use an inline comment instead of `{{content-for 'body'}}`?

Maybe something like:

```javascript
if (type === 'body') {
  return "// EMBER_CLI_FASTBOOT_BODY";
}
```

Then tweak the replacement to look for that string instead?

---

I find that *much* easier to grok (and doesn't overload the concept of `{{content-for 'body'}}` like ember-cli-fastboot does today).

gdub22: @rwjblue good idea.  Also, it will have to be an html comment otherwise that string is gonna show up on the page in a standard ember serve (non-fastboot)

rwjblue:
Sorry for wasting your time here, I was blocked on this so I submitted https://github.com/tildeio/ember-cli-fastboot/pull/14 (which is based on your commit) and made the changes mentioned above.

PR #14, opened 2015-04-08
Honor other addon's content-for hooks.

rwjblue:
Supercedes #13.


gdub22:
@rwjblue :+1: thanks!
Note that this still first looks for 'app/fastboot.html' in it's raw unbuilt form, if it exists. (I wasn't sure the use case for it, so I left it as-is).  So you'll still have to strip the content-for blocks out of there.

PR #20, opened 2015-04-08
Add command line option for serving assets.

rwjblue:
This takes @hone's commits from #6 and supercedes it.

You can run serving assets via:

```javascript
ember fastboot --serve-assets
```

---

Please note: that pre-Glimmer Ember will not be able to start the application where the server rendered data left off.  Currently, this would mean that when the Ember app is booted in the client it will be duplicated in the DOM. To work around this, I have added an instance-initializer that clears the rootElement before initial render.


rwjblue:
Thanks to @hone for doing all the hard work on this one....

PR #22, opened 2015-04-23
Add title handling.

rwjblue:
Adds support for addons/apps to update the document title by using `renderer:-dom`.


rwjblue:
Tested and works very well with ember-cli-document-title@0.1.0.

PR #23, opened 2015-04-23
Allow preloading of data looked up during fastboot rendering.

rwjblue:
Using the provided store will seed the store in the booted application with the data that was fetched during the Fastboot rendering process.

---

In your app you need to use `ember-cli-fastboot/store` as your stores base class. Often this just means importing and reexporting in your `app/store.js` file:

```javascript
// app/store.js
import Store from 'ember-cli-fastboot/store';

export default Store;
```

### DIFF
--- a/addon/instance-initializers/preload-data.js
+++ b/addon/instance-initializers/preload-data.js
@@ -1,0 +1,30 @@
+import Ember from 'ember';
+
+const { $, keys } = Ember;
+
+export default {
+  initialize(instance) {
+    if (!$) {
+      return;
+    }
+
+    const rawDump = Ember.$('meta[name="preload-data"]').attr('content');
+
+    if (!rawDump) {
+      return;
+    }
+
+    const data = JSON.parse(decodeURIComponent(rawDump));
+    const storeNames = keys(data);
+
+    for (let index = 0, length = storeNames.length; index < length; index++) {
+      const storeFullName = storeNames[index];
+      const store = instance.container.lookup(storeFullName);
+      const storeData = data[storeFullName];
+
+      if (store && store.populateCache) {
+        store.populateCache(storeData);
+      }
+    }
+  }
+};
--- a/addon/store.js
+++ b/addon/store.js
@@ -1,0 +1,48 @@
+import Ember from 'ember';
+import DS from 'ember-data';
+
+const keys = Ember.keys;
+const create = Ember.create;
+export const knownStores = [];
+
+export default DS.Store.extend({
+  init() {
+    this._super(...arguments);
+    knownStores.push(this._debugContainerKey);
+  },
+
+  populateCache(dump) {
+    const types = keys(dump);
+
+    for (let index = 0, length = types.length; index < length; index++) {
+      const typeKey = types[index];
+      const records = dump[typeKey];
+
+      this.pushMany(typeKey, records);
+    }
+  },
+
+  serializeCache() {
+    let results = create(null);
+
+    const types = keys(this.typeMaps);
+
+    for (let index = 0, length = types.length; index < length; index++) {
+      const typeId = types[index];
+      const typeKey = this.typeMaps[typeId].type.typeKey;
+      const records = this.typeMaps[typeId].records;
+      const recordCount = records.length;
+
+      results[typeKey] = [];
+
+      for (let currentRecord = 0; currentRecord < recordCount; currentRecord++) {
+        const record = records[currentRecord];
+        const serializedRecord = record.toJSON({ includeId: true });
+
+        results[typeKey].push(serializedRecord);
+      }
+    }
+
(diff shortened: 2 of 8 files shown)


app/initializers/fastboot.js
```diff
@@ -35,7 +35,13 @@ export default {
         return App.visit(url).then(function(instance) {
           var view = instance.view;
           var title = view.renderer._dom.document.title;
-          var element;
+          var store = view.container.lookup('store:main');
+          var element, data;
+
+          if (store.serializeCache) {
+            console.log('serializing');
```
stefanpenner: ^

addon/store.js
```diff
@@ -0,0 +1,40 @@
...
+      const records = dump[typeKey];
+
+      this.pushMany(typeKey, records);
+    }
+  },
+
+  serializeCache: function() {
+    let results = {};
+
+    for (let typeId in this.typeMaps) {
```
stefanpenner: `Ember.keys`

addon/store.js
```diff
@@ -0,0 +1,40 @@
...
+      }
+
+      const records = dump[typeKey];
+
+      this.pushMany(typeKey, records);
+    }
+  },
+
+  serializeCache: function() {
+    let results = {};
```
stefanpenner: maybe a null prototype Object?

app/initializers/fastboot.js
rwjblue: Whoops, good catch.  Fixed.

addon/store.js
rwjblue: :+1:

addon/store.js
rwjblue: :+1: - Updated

rwjblue:
Updated per review.

tchak:
What happens if an app use multiple stores?

rwjblue:
Ember data itself assumes a single store and uses store:main, which is why I used it here. I'd be happy to implement multi store caching if multiple stores are indeed a "supported" and required thing here, but we will have to figure out a way to discover the store location (as you mentioned).

This pull request addresses the significant majority of cases where only a single store is present.

tchak:
Ember data do support multiple stores and some people (my self included) use them in production apps. I am happy with this PR as first step. Maybe we can just use named metas. At least on loading we could avoid hardcoded store name. In the part that serialize store hardcoding `store:main` seems fine for now.

stefanpenner:
> Ember data do support multiple stores and some people (my self included) use them in production apps. I am happy with this PR as first step. Maybe we can just use named metas. At least on loading we could avoid hardcoded store name. In the part that serialize store hardcoding store:main seems fine for now.

how about we just key it 'store:main' and allow multiple store keys, as long as they are resolvable the above code could look them up and insert the payload.

tchak:
@stefanpenner this is what I was trying to say :) Sorry it was not very clear. :+1:

rwjblue:
Yep, I understood.  Working on the changes now..

rwjblue:
Updated, now no longer tied to any hard coded store names.

stefanpenner:
> Updated, now no longer tied to any hard coded store names (or limited to usage of a single store).
>   Merge pull request

nice!

addon/instance-initializers/preload-data.js
```diff
@@ -0,0 +1,30 @@
...
+
+const { $, keys } = Ember;
+
+export default {
+  initialize(instance) {
+    if (!$) {
+      return;
+    }
+
+    const rawDump = Ember.$('meta[name="preload-data"]').attr('content');
```
igorT: Shouldn't this be just `$`?

addon/store.js
```diff
@@ -0,0 +1,48 @@
...
+import DS from 'ember-data';
+
+const keys = Ember.keys;
+const create = Ember.create;
+export const knownStores = [];
+
+export default DS.Store.extend({
+  init() {
+    this._super(...arguments);
+    knownStores.push(this._debugContainerKey);
```
igorT: If this is run on the server, wouldn't you expect to init a particular store many times even though you import it once? Probably need to make sure the same key is not pushed multiple times

addon/instance-initializers/preload-data.js
rwjblue: Yes, good catch.

addon/store.js
rwjblue: A given `store:<name>` should only be looked up once (it is a singleton in the container).  Each request gets a new container instance.  So each store (by name) should only be instantiated once.

igorT: But you are closing over an array that is static? Doesn't the knownStores array leak between requests?

rwjblue: And this is why I love working with smart people.  Good catch, will fix.

PR #32, opened 2015-05-20
Send responses in chunks, flushing between static and dynamic sections.

rwjblue:
This should allow the bulk of the `<head>` content to be sent before the Ember app is booted and rendered in fastboot.  Thus allowing the client browser to start fetching assets and whatnot.

A similar technique is done in various other frameworks (including Rails).

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,7 +11,7 @@ module.exports = {
     };
   },
 
-  contentFor: function(type, config) {
+  contentFor: function(type) {
     // do nothing unless running `ember fastboot` command
     if (!process.env.EMBER_CLI_FASTBOOT) { return; }
 
@@ -19,7 +19,7 @@ module.exports = {
       return "<!-- EMBER_CLI_FASTBOOT_BODY -->";
     }
 
-    if (type === 'head') {
+    if (type === 'head-footer') {
       return "<!-- EMBER_CLI_FASTBOOT_TITLE -->";
     }
 
--- a/lib/models/server.js
+++ b/lib/models/server.js
@@ -9,7 +9,7 @@ function FastBootServer(options) {
     vendorFile: options.vendorFile
   });
 
-  this.html = fs.readFileSync(options.htmlFile, 'utf8');
+  this.htmlChunks = this.getChunks(fs.readFileSync(options.htmlFile, 'utf8'));
 
   this.ui = options.ui;
 }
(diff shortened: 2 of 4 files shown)


bower.json
```diff
@@ -2,7 +2,7 @@
   "name": "ember-cli-fastboot",
   "dependencies": {
     "jquery": "^1.11.1",
-    "ember": "canary",
+    "ember": "components/ember#8fababb25f6aa7c0a172bb2e3020c8fd63a8e4c1",
```
rwjblue: This is locked down to the SHA just before Glimmer landing (since Fastboot support isn't ready there yet).

stefanpenner:
:+1:

chancancode:
I'm currently working on (at least *thinking about*) redirects: the basic idea is that if the ember app ended up on a different URL than it started with, then the fastboot server should send a redirect instead. If we eventually need to support that (and I am pretty sure we must), we will loose part of this benefit – specifically "allow content to be sent before the Ember app is booted **and rendered** in fastboot", because we don't know what headers (status code and `Location`) to send before the app settles (the current implementation in the PR doesn't try to take advantage of that anyway).

Given that, does flushing in chucks still have merits on its own? (I don't know enough about node and http to know :smile:)

chadhietala:
I'm wondering if more granular chunking could be done. Basically when the `afterRender` queue is about to be flushed you return the HTML. This would mean that `App.visit` returns an observable instead of a promise.

chancancode:
Sorry, forgot to update. Discussed more with @rwjblue about this. While the *original* intention of is to flush the `<!DOCTYPE><html>....</head><body>` while we wait for the app to boot and render. As I mentioned above (and as @rwjblue independently realized while working on this), this does not work in practice because it forces us to also flush the HTTP headers – among other things, we don't know what status code to send yet (200 vs 404 vs 500 vs 302...).

So the current implementation (and likely all we can do) is to still wait for the app to boot/render, but once we determined that we are going to render a 200, we flush once between `<!-- EMBER_CLI_FASTBOOT_BODY -->` and the rest of the body (the app HTML, which could potentially be large). Out of the box, this should allow (in-theory) the browser's parser to notice your CSS and start pre-fetching them while downloading the rest of the body. If you move your `<script>` tag to before `{{content-for 'body'}}`, that should allow the browser to pre-fetch your scripts while downloading the body.

I'm conscious of the complexity this will add. The current implementation is pretty simple, but it seems like it will generally make things more complicated down the road. For example, [in Rails](https://github.com/rails/rails/blob/b6760d8f14af6b00250395f2c830b89e43621860/actionpack/lib/action_controller/metal/streaming.rb), this is disabled by default and there are plenty of caveats mentioned in the documentation there. To be fair, most of these problems can probably be attributed to 1) they are doing what this PR aims to do originally (stream while rendering), and 2) the Ruby Rack API is not originally designed for streaming. Neither of these are actually a problem here, but it made me feel more cautious about the potential hidden cost down the road.

I'm also don't know if merely flushing between `<head>` and `<body>` in-and-of-itself makes much of a different at all – most references (including the Rails documentation I linked and other node resources) seems to suggest that only really beneficial if you can flush the assets *before* doing the heavy lifting (rendering).

I did my own investigation, and it seems to confirm that this basically makes no difference.

Script: https://gist.github.com/chancancode/4884a321eb29b7dbcd42 (I based the file sizes roughly on this current page, which imo is on the "very large" side of things.)

I tested this on chrome, using the network emulation to simulate 2G situation and disabled cache. The difference basically falls within the S.D. (~5ms out of 10.5s total). I believe that when you `send` a large body back at once, the response is essentially still "streamed" but at the TCP layer, and the browser's network stack/parser basically end up doing identical things here. So basically, the chunking is only helpful if you have work that you need to do in between chunks, since we don't have deferrable work in the end, it basically defeats the purpose. If someone can help prove me wrong that would actually be very helpful :)

stefanpenner:
No you are correct. Flushing only really matters if you have to wait for some subsequent data to be processed or fetched. Otherwise the existing system handles it for you.

I am more interested in flushing vendor and style parts of head before the ember app renders. So the sub requests they spawn can start asap. After that, re-ordering the content html and the preloaded data. So the content is visible before the preloaded data streams in.

These two combined would help the dockyard blog case, where initial visual html is postponed due to being placed after the preloaded data blob.

PR #33, opened 2015-05-21
Actually return 404 status code

### LINKED ISSUE #31: Return 404 for unrecognized route?
https://github.com/tildeio/ember-cli-fastboot/blob/f722fa69077ce83b9153c363059de09823adcd18%5E/lib/fastboot-server.js#L29

Seems like this should be a 404 :)

chancancode:
Fixes #31, since no one claimed it yet :stuck_out_tongue: 

As far as I can tell this is just a typo (introduced in 3c0a499), there doesn't seem to be a real reason to return 400 instead of 404.


rwjblue:
Awesome, thanks!

chancancode:
@rwjblue by the way, how do you actually run the acceptance tests? `ember test` gave me "can't find module chai" (among 1000 jshint errors). If you can help me figure that out, I can add a test for this (I am also experimenting with another thing).

PR #37, opened 2015-05-25
Read asset paths from EmberApp outputPaths option

### LINKED ISSUE #36: Doesn't read from custom EmberApp outputPaths
Fastboot currently assumes your `EmberApp` uses the stock [outputPaths](http://www.ember-cli.com/#configuring-output-paths) for its assets.  

https://github.com/tildeio/ember-cli-fastboot/blob/master/lib/commands/fastboot.js#L81-L91

gdub22:
Fixes #36


lib/commands/fastboot.js
```diff
@@ -64,9 +64,8 @@ module.exports = {
     this.commandOptions = options;
 
     var runCommand = function() {
-      var appName = process.env.EMBER_CLI_FASTBOOT_APP_NAME || this.project.name();
-
-      return this.runCommand(appName, options);
+      var appOutputPaths = JSON.parse(process.env.EMBER_CLI_FASTBOOT_APP_OUTPUT_PATHS);
```
rwjblue: This needs to have a fallback. Otherwise fastboot can only be used while building (since included hook is only ran for builds).

Maybe we should add a fastboot specific config file at config/fastboot.js for the shared config?

gdub22: @rwjblue not sure how to proceed here. Since the app has to be built at least once to serve it, write the output paths to a temp file?

PR #38, opened 2015-05-25
Fix clear-double-boot with virtual application view

gdub22:
The temporary [initializer](https://github.com/tildeio/ember-cli-fastboot/commit/1a71c5f2843b81bb8875abb54a9561dac27a6f94) to remove the double-rendered app when fastbooting with `--serve-assets` was not working when the application view is virtual.

### DIFF
--- a/app/instance-initializers/clear-double-boot.js
+++ b/app/instance-initializers/clear-double-boot.js
@@ -12,7 +12,7 @@ export default {
     var originalDidCreateRootView = instance.didCreateRootView;
 
     instance.didCreateRootView = function() {
-      Ember.$(instance.rootElement + ' .ember-view').remove();
+      Ember.$(instance.rootElement).children().not('script').remove();
 
       originalDidCreateRootView.apply(instance, arguments);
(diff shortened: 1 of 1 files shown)


rwjblue:
The application view is not virtual by default, are you setting its tagName manually?

rwjblue:
It is fairly common to have other content inside the rootElement (just like you saw with the script tags). This would remove that content...

gdub22:
I usually set `tagName: ''` to avoid the extra wrapper element.  I see your point that there could be other static content that would get removed.  I can live with the default non-virtual view for now.

rwjblue:
@gdub22 - Another option for your particular use case, would be to make an initializer in your app that does this (that is what we did before adding it here).

Something like:

```javscript
// app/initializers/clear-double-boot.js
import config from '../config/environment';

export default {
  name: 'clear-double-boot',
  initializer: function() {
    Ember.$(config.APP.rootElement).children().not('script').remove();
  }
}
```

PR #41, opened 2015-05-27
Include the --serve-assets flag in README

kellysutton:
This tripped me up the first few times I tried to work with FastBoot. I propose it be included in the README, be set as a default, or we leave a note as to why neither of the first two options were taken.


rwjblue:
Seems good to me!

PR #49, opened 2015-07-02
Implement naive meta tag support

jpadilla:
After hearing about Dockyard's new site using fastboot, we wanted to take it out for a spin on a new project. We noticed it wasn't yet possible to set meta tags like `description`. This allows setting meta tags similar to how we can set the document's title.

```javascript
Ember.set(renderer, '_dom.document.title', 'Application Route -- Title');
Ember.set(renderer, 'metaTags', [{
  name: 'description',
  content: 'something here'
}, {
  property: 'og:image',
  content: 'http://placehold.it/500x500'
}]);
```

This obviously still feels hacky but seems to work just fine so we decided to share and hopefully get some feedback.

Update: With a few simple tweaks this and #23 can play nicely together.

### DIFF
--- a/app/initializers/fastboot.js
+++ b/app/initializers/fastboot.js
@@ -34,23 +34,47 @@ export default {
         FastBoot.debug("routing; url=%s", url);
 
         return App.visit(url).then(function(instance) {
+          var serializer = new SimpleDOM.HTMLSerializer(SimpleDOM.voidMap);
           var view = instance.view;
-          var title = view.renderer._dom.document.title;
+          var dom = view.renderer._dom;
+          var title = dom.document.title;
+          var head = dom.document.head;
           var metaTags = view.renderer.metaTags;
+          var body;
           var element;
 
           Ember.run(function() {
-            element = view.renderToElement();
+            body = view.renderToElement();
           });
 
-          element = element.firstChild;
+          if (title) {
+            element = dom.document.createElement('title');
+            element.appendChild(dom.document.createTextNode(title));
+            head.appendChild(element);
+          }
 
-          var serializer = new SimpleDOM.HTMLSerializer(SimpleDOM.voidMap);
+          if (metaTags) {
+            metaTags.forEach(function(metaTag) {
+              element = dom.document.createElement(metaTag.name);
+
+              Object.keys(metaTag.attrs).forEach(function(key) {
+                element.setAttribute(key, metaTag.attrs[key]);
+              });
+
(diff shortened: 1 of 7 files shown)


gdub22:
@jpadilla  :heart: Looks great.  Some ideas for additional support:

- Supporting metadata with other tags such as `link` (canonical, author, etc).
- Exposing the api so the Ember router can keep things in sync during clien-side transitions.

gcollazo:
@gdub22 :+1: for supporting supporting `<link>` tags.

What is the use case for updating `<meta>` or `<link>` tags while the app is running in the browser? I believe those are only important when the page is requested by a crawler in which case fastboot would do the right thing.

gdub22:
@gcollazo you are right about the crawler, it would just be nice to keep them in sync.  Thinking about it more, it would be good to have this as an addon where you could define meta for each route. (https://github.com/ronco/ember-cli-meta-tags is a start) and fastboot just provides an injection point.  Kinda how the title works now with https://github.com/kimroen/ember-cli-document-title

ronco:
This is great idea.  I'd love to see it support arbitrary head content.

jasonmit:
> This is great idea. I'd love to see it support arbitrary head content.

I second this.  I'll need to be able to inject additional scripts into the head dynamically and would love for this to be generic enough to support all head content.

rwjblue:
I would prefer tweaking [simple-dom here](https://github.com/krisselden/simple-dom/blob/master/lib/simple-dom/document.js#L11) to add:

```javascript
this.head = new Element('head');
```

Then we can handle arbitrary head content here....

jpadilla:
@rwjblue btw this is ready for another look whenever you get the chance.

jpadilla:
Tests are currently not passing, since we'd need https://github.com/krisselden/simple-dom/pull/12 to be merged.

gdub22:
@jpadilla :+1: I'd say that the head addition by itself is enough, then let the developer/addons implement the meta tags/arbitrary head content.  I have this [addon](https://github.com/gdub22/ember-document-meta) just about working with your additions here and in simple dom.

jasonmit:
@jpadilla Any progress?  I can help out if it's a time thing.

jpadilla:
@jasonmit feel free to pick up. There's probably just a few things left after the simple-dom update.

esistgut:
Are there updates on this?

ronco:
@esistgut I was actually planning on taking a crack at finishing it up this evening.

tomdale:
Closed in favor of #114

PR #59, opened 2015-08-14
Update installation instructions, removing unnecessary references to …

kwight:
…`noneLocation` and `HTMLbars`.

From what I can tell, the change to `router.js` is not required, and HTMLbars has been in the release build for quite some time.


rwjblue:
thank you!

Issue #46, opened 2015-06-09
Ember FastBoot running at http://:::3000


furley:
When I run fastboot, it seems to run without a host, (http://:::3000 rather than http://127.0.0.1:3000). Would that issue be with my local environment, or with fastboot? Or is that not an issue at all?

```
DEBUG: -------------------------------
DEBUG: Ember      : 1.12.1
DEBUG: Ember Data : 1.0.0-beta.20+canary.60c01a3f30
DEBUG: -------------------------------
Ember FastBoot running at http://:::3000
```

felixbuenemann:
That's the short form for all zeroes in ipv6, meaning it’s bound to all interfaces. The proper display for ipv6 in urls should be http://[::]:3000 which is the equivalent of http://0.0.0.0:3000 in ipv4. To access localhost via ipv6 you can use http://[::1]:3000.

tomdale:
This is listening to all interfaces instead of just localhost. We should definitely make this configurable, particularly for production. Tracked in #58.

felixbuenemann:
@tomdale There's also a display issue for the ipv6 case. According to RFC 3986 section 3.2.2 ipv6 addresses in URIs should be enclosed in square brackets. So the proper URI would be `http://[::]:3000`.

tomdale:
@felixbuenemann I'm not sure if I'll have time to get to that right away, but I would gladly accept a PR. The relevant code is here: https://github.com/tildeio/ember-cli-fastboot/blob/master/lib/commands/fastboot.js#L40-L43

felixbuenemann:
@tomdale Done, see PR #60.

PR #60, opened 2015-08-14
Fix display of ipv6 fastboot server uri

felixbuenemann:
IPv6 addresses must be enclosed in square brackets when used inside URIs to avoid ambiguity when parsing the port (see RFC 3986 section 3.2.2).

This addresses the URI display issue discussed in #46.


tomdale:
Thanks!

PR #70, opened 2015-10-08
use new BootOptions

victor-stone:
Not 100% sure if I did this right but it works for the latest canary

### DIFF
--- a/app/initializers/fastboot.js
+++ b/app/initializers/fastboot.js
@@ -40,7 +40,10 @@ export default {
 
         var promise;
         Ember.run(function() {
-          promise = App.visit(url);
+          promise = App.visit(url {
+            document: doc,
+            rootElement: doc
+          });
         });
 
         return promise.then(function(instance) {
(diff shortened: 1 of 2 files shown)


app/initializers/fastboot.js
```diff
@@ -40,7 +40,10 @@ export default {
 
         var promise;
         Ember.run(function() {
-          promise = App.visit(url);
+          promise = App.visit(url {
```
topaxi: Missing comma :)

rwjblue:
@chancancode - r?

chancancode:
I'm working on a more complete patch: https://github.com/tildeio/ember-cli-fastboot/pull/71

I'll be on a plane soon, so I pushed it up so you can start poking at it. (It might or might not work, but that's the broad strokes)

rwjblue:
Thanks @chancancode!

@victor-stone - I'm going to close this in favor of #71.  We can always revisit (pun intended) if #71 ends up taking too long...

PR #71, opened 2015-10-08
Use latest visit API

chancancode:
Update fastboot server for https://github.com/emberjs/ember.js/pull/12394

Require coordination from https://github.com/ember-fastboot/ember-fastboot-server/pull/2

- [x] Take advantage of the new API
- [x] FastFail™
- [x] Don't mutate `autoBoot`
- [x] Remove `FastBoot` global
- [x] Better integration with Ember CLI
- [x] Move most of the work out of initializers
- [x] Test that it works
- [ ] Write tests

Open question:

* What are the rules regarding passing complex objects across the sandbox boundary? It would not be hard to avoid that if we have to.

cc @tomdale @stefanpenner


index.js
```diff
@@ -11,6 +28,22 @@ module.exports = {
...
 
+  config: function(/* environment, appConfig */) {
+    // do nothing unless running `ember fastboot` command
+    if (!process.env.EMBER_CLI_FASTBOOT) { return {}; }
+
+    return {
+      autoRun: false,
+      storeConfigInMeta: false,
+      EmberENV: {
+        FEATURES: { 'ember-application-visit': true }
```
rwjblue: I initially avoided doing this because feature flags were not being merged.  Which meant that if you had a feature flag set in the consuming apps `config/environment.js` it would completely ignore any features added from an addon's `config` hook.

This may be fixed now with current ember-cli, but I'd love it if you could confirm.  The easiest way to test would be to make a throw away app, set some features in it, run `ember fastboot` and confirm that the values are merged propertly in the resulting files...

index.js
```diff
@@ -11,6 +28,22 @@ module.exports = {
...
+      },
+      APP: {
+        autoBoot: false
```
rwjblue: Similar concern to the above.  I would love to be wrong about it though.

app/initializers/fastboot.js
```diff
@@ -4,62 +4,18 @@ export default {
...
 
   initialize: function(App) {
-    // Detect if we're running in Node. If not, there's nothing to do.
-    if (typeof document === 'undefined') {
-      var doc = new SimpleDOM.Document();
-      var domHelper = new Ember.HTMLBars.DOMHelper(doc);
-
-      domHelper.protocolForURL = function(url) {
+    // Detect if we're running in FastBoot mode. If not, there's nothing to do.
+    if (FastBoot) {
```
stefanpenner: should this be typeof check? Or will the ambient global always be present if this initializer runs?

index.js
stefanpenner: :+1:

index.js
```diff
@@ -23,17 +56,8 @@ module.exports = {
...
     }
 
-    if (type === 'vendor-prefix') {
-      return '// Added from ember-cli-fastboot \n' +
-             'EmberENV.FEATURES = EmberENV.FEATURES || {};\n' +
-             'EmberENV.FEATURES["ember-application-visit"] = true;\n';
-    }
-  },
-
-  included: function() {
```
rwjblue: :+1: avoiding `included` hook allows us to have fastboot be available in nested addons (nested addons normally do not have `included` hook called on them).

app/initializers/fastboot.js
rwjblue: @stefanpenner Good point (currently this initializer is run in both fastboot and non-fastboot modes), on certain browsers (looking at you IE) attempting to access an unprefixed global (without using `typeof`) results in an error...

As an aside, we could easily make the `treeForApp` hook a noop if not running the `ember fastboot` command (which would mean apps do not pay any filesize penalty when using `ember build` or `ember s`).

stefanpenner: > certain browsers

Isn't this the same across all browsers?

chancancode: Hmm treeForApp is an interesting.

I can also domnake it a typeof check.

index.js
chancancode: Hmm, from what I read, the whole thing is deep merged with lodash. I will confirm!

index.js
```diff
@@ -23,17 +65,28 @@ module.exports = {
...
+      });
+    } else {
+      return new Funnel(tree, {
+        annotation: 'Funnel: Remove server-only initializers',
+        exclude: [
+          'initializers/server/*',
+          'instance-initializers/server/*'
+        ]
+      });
     }
```
chancancode: @rwjblue @stefanpenner how does that look? with this I could complete remove the `FastBoot` global we inject into the sandbox.

By the way, if this works out for us, we might want to move it into the `preprocessTree` hook so that apps and other addons can use the same pattern

rwjblue: :+1: - This looks great to me!

---

I don't see a way to completely generalize this, but we can chat about that further in a separate issue...

chancancode: Wouldn't it just work if we move this code to `preprocessTreeFor`? Any initializers in with a `browser/...` prefix will be stripped for server builds and vice versa. Maybe I am understanding it incorrectly – does `preprocessTreeFor` not get the fully merged app tree with user code and addon code all merged together?

rwjblue: @chancancode - Ya, we could come up with something that could work (I misread initially as `postprocessTree` not `preprocessTree`).  I'm happy to come up with a specific layout/plan to make it a standard thing.

chancancode:
Got the existing tests passing today with some minor fixes! I'll clean it up tomorrow!

index.js
```diff
@@ -11,6 +37,25 @@ module.exports = {
     };
   },
 
+  included: function(app) {
+    app.options.autoRun = false;
+    app.options.storeConfigInMeta = false;
+  },
```
chancancode: @rwjblue turns out removing this is not as easy as I thought – contrary to the [documentation](https://github.com/ember-cli/ember-cli/blob/master/ADDON_HOOKS.md#config) (see "Uses"), these options can't actually be set through the `config` hook :( (It makes sense, mutating these *build options* at a random time during build probably has the same problem as setting `autoboot` during boot :wink:)

@stefanpenner can you c/d? If that's true I can (eventually) send a doc PR

index.js
```diff
@@ -11,6 +37,25 @@ module.exports = {
...
+    // do nothing unless running `ember fastboot` command
+    if (!process.env.EMBER_CLI_FASTBOOT) { return {}; }
+
+    return {
+      EmberENV: {
+        FEATURES: { 'ember-application-visit': true }
+      },
+      APP: {
+        autoboot: false
+      }
```
chancancode: both of these works (they merges correctly as far as I can tell, can add a test too)

index.js
chancancode: I'll leave that for another time, probably until someone asks for it :P (if you are reading this, you might just want to PR it :wink:)

package.json
```diff
@@ -18,20 +18,20 @@
...
+    "ember-cli-content-security-policy": "0.4.0",
+    "ember-cli-dependency-checker": "^1.0.1",
+    "ember-cli-htmlbars": "0.7.9",
+    "ember-cli-htmlbars-inline-precompile": "^0.2.0",
+    "ember-cli-ic-ajax": "0.2.1",
+    "ember-cli-inject-live-reload": "^1.3.1",
+    "ember-cli-qunit": "^1.0.0",
+    "ember-cli-uglify": "^1.2.0",
+    "ember-data": "1.13.8",
+    "ember-export-application-global": "^1.0.3",
```
chancancode: I was debugging something unrelated, so I'll need to double check that the upgrade is actually necessary

tests/acceptance/simple-test.js
```diff
@@ -30,6 +30,9 @@ describe('simple acceptance', function() {
         expect(response.headers["content-type"]).to.eq("text/html; charset=utf-8");
         expect(response.body).to.contain('<title>Application Route -- Title</title>');
         expect(response.body).to.contain("Welcome to Ember.js");
+      }, function(error) {
+        console.log(error.stack);
+        throw error;
       });
```
chancancode: Oops, will rm

chancancode:
Tests are green, fixed all the problems, backfilled the commit messages, did some smoke tests :grimacing: 

@tomdale @rwjblue final review?

If this looks okay to you, we should probably first merge ember-fastboot/ember-fastboot-server#2, release 0.1.0, then update the last commit to use the released version before merging.

tests/helpers/start-server.js
```diff
@@ -36,7 +37,10 @@ module.exports = function runServer(callback, options) {
   args.push(commandOptions);
 
   return new RSVP.Promise(function(resolve, reject) {
-    runCommand.apply(null, args)
+    runCommand.call(null, emberCommand, 'build') // build 'dist'
```
chancancode: So this reflects a real-world problem: you need to now run `ember build` (probably `ember build -prod`) before you run `ember fastboot --serve-assets`, because we need to serve the browser builds from `dist`. (See the commit message for details)

Unfortunately, it is not easy to refactor the set up to produce both builds in a single command (it relies on a global ENV variable, among other things).

So I just added a new `assets-path` option for now, but we should revisit this in the future.

tomdale:
Released 0.1.0 of `ember-fastboot-server`, I will merge this and update the dependency.

PR #72, opened 2015-10-09
End-to-end smoke tests

tomdale:
The basic idea here is to borrow from what Ember CLI does for its own suite of tests. Namely, we want to do an end-to-end smoke test that creates a new Ember application (via `ember new`) and then goes through all of the steps needed to make an Ember CLI app work with FastBoot. Once that's done, we boot up the server and make sure that runs correctly.

In the interests of performance, you don't want each test app to `npm install` (which takes ages), so there's some complexity in this test around only doing one `npm install` and `bower install` and then re-using the resulting `node_modules`/`bower_components` across runs. Because even one `npm install` is agonizingly slow when running the tests, I also provided the ability to provide a "pre-cooked" `node_modules` directory that can be used across multiple runs.


tomdale:
@rwjblue @chancancode This is ready for your review.

tomdale:
@rwjblue @chancancode The messiness of the inherited code from Ember CLI was bugging me, so I rewrote it today. It's much cleaner, faster, and more robust. Recommended workflow is:

1. `npm run test:precook` to save `node_modules`
2. `DEBUG=fastboot-test npm test` to run the tests with verbose logging turned on.

lib/ext/promise.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
```
rwjblue: This can likely be removed.  Looks like it is only required in two places:

* tests/helpers/acceptance.js
* tests/helpers/run-command.js

In both cases, I think `var Promise = require('rsvp').Promise;` should work.

tests/precook-node-modules.js
```diff
@@ -0,0 +1,24 @@
+"use strict";
+
+var temp = require('temp').track();
+var runCommand       = require('./helpers/run-command');
```
rwjblue: Odd spacing here

tests/precook-node-modules.js
```diff
@@ -0,0 +1,24 @@
...
+var path = require('path');
+
+var tmpDir = temp.mkdirSync();
+var root = process.cwd();
+var moveDirectory    = require('./helpers/move-directory');
```
rwjblue: Seems like this file is most just 1 space (not space aligned), can this extra space be removed?

tests/runner.js
```diff
@@ -0,0 +1,34 @@
...
+  timeout: 5000,
+  reporter: 'spec'
+});
+
+var testFiles = glob.sync(root + '/**/*-test.js');
+
+addFiles(mocha, testFiles);
+
+try {
+    runMocha();
```
rwjblue: Too much whitespace here

rwjblue:
This looks great!  All of my legit concerns have been addressed, only silly whitespace comments left (but feel free to ignore).

:+1:

searls:
:+1: this looks awesome and like just as broad of coverage you'd ideally want to see in a tool like this. Smart to prevent duplicative npm/bower installs.

rwjblue:
:shipit:

chancancode:
:metal:



—

PR #76, opened 2015-11-02
Update README.md

olivierlesnicki:
Both ember and ember-data needs to point to the canary branch since ember v2.x.x is incompatible with ember-data v1.x.x

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ In order to get FastBoot working, you will first need to enable Ember Canary:
 
 ```
 rm -rf bower_components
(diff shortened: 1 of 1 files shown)


rwjblue:
Ember Data canary is not needed.

rwjblue:
Must be Ember Data 2.x (which is available as `~2.0.0`, `~2.1.0`, `release`, `beta`, or `canary`)

tomdale:
Thanks for the PR, closing because I think it is out of date. Please reopen if I'm mistaken. Thanks!

PR #78, opened 2015-12-17
Better failure detection

tomdale:
This commit wraps each file that gets put in `vendor` in a `try/catch`. This allows us to capture FastBoot-incompatible addons and print a much more useful error message in the console.

Note that this relies on monkeypatching Ember CLI's `EmberApp` method in a very hacky way. @stefanpenner thinks it may be appropriate to expose a hook that we can use instead.

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,12 +42,11 @@ module.exports = {
     if (process.env.EMBER_CLI_FASTBOOT) {
       process.env.EMBER_CLI_FASTBOOT_APP_NAME = app.name;
       app.options.autoRun = false;
+      patchEmberApp(app);
     }
 
     // We serve the index.html from fastboot-dist, so this has to apply to both builds
     app.options.storeConfigInMeta = false;
-
-    patchEmberApp(app);
   },
 
   config: function(/* environment, appConfig */) {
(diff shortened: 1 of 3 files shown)


rwjblue:
Doesn't this break source maps?

rwjblue:
I am also wondering about performance. It is my understanding that using a try catch in some cases prevent V8 from optimizing. Is that an issue here?

stefanpenner:
> Doesn't this break source maps?

Yes it will, we will have to come up with a better strategy at some point. That being said, if this is only for fast-boot builds it may be ok for awhile?

stefanpenner:
@rwjblue it would prevent the context from being optimized, but not nested contexts. I am unsure as to the magnitude of the issue.

There is an alternative way for us to do this without try/catch, if it turns out to be a measurable issue.

```
function checkDeps() {
setTimeout(function() { /* check which deps have run and which haven't figuring out what went sideways */ });
}

startScript('jQuery');
//..
endScript('jQuery');

```

tomdale:
I think https://github.com/ember-fastboot/ember-fastboot-server/pull/16 is a better solution to this. Let's revisit if this remains a problem for people.

PR #81, opened 2015-12-21
Enable `storeConfigInMeta` only for FastBoot builds

### LINKED ISSUE #80: Wrong env in /tests while running `ember server`
Symptom:

1. `http://localhost:4200/tests` is incorrectly in `development` mode while running `ember server`
1. `http://localhost:4200/tests` is correctly in `test` mode while running `ember test --server`

If `ember-cli-fastboot` is removed from `package.json`, the erroneous behavior disappear.

```
$ ember -v
version: 2.2.0-beta.1
node: 0.12.1
npm: 2.14.10
os: darwin x64
```

zeppelin:
Fixes #80

### DIFF
--- a/index.js
+++ b/index.js
@@ -41,10 +41,10 @@ module.exports = {
     if (process.env.EMBER_CLI_FASTBOOT) {
       process.env.EMBER_CLI_FASTBOOT_APP_NAME = app.name;
       app.options.autoRun = false;
+      
+      // We serve the index.html from fastboot-dist, so this has to apply to both builds
+      app.options.storeConfigInMeta = false;
     }
-
-    // We serve the index.html from fastboot-dist, so this has to apply to both builds
-    app.options.storeConfigInMeta = false;
   },
 
   config: function(/* environment, appConfig */) {


rwjblue:
:+1:

index.js
```diff
@@ -41,10 +41,10 @@ module.exports = {
     if (process.env.EMBER_CLI_FASTBOOT) {
       process.env.EMBER_CLI_FASTBOOT_APP_NAME = app.name;
       app.options.autoRun = false;
+      
+      // We serve the index.html from fastboot-dist, so this has to apply to both builds
+      app.options.storeConfigInMeta = false;
```
stefanpenner: @tomdale i would prefer if this was loaded different, but still external to the app bundle. Otherwise we will pwn the "binary identical app.js" thing we spoke about at the last meeting.

It might be fine for now, but this should be added to the checklist of things to fix/think about.

krisselden: We solve this by having config/environment branch, on the server side looking for it on the sandbox and on the client side looking for it in meta.  I agree with @stefanpenner we should not be  building app.js twice once for the server and once for the client.

stefanpenner: ideally it would be something like

```js
  bootApp(require('app/app'), loadConfig(...);
```
or something

nathanhammond:
See also:
* https://github.com/ember-cli/ember-cli/pull/5246
* https://github.com/ember-cli/rfcs/pull/30

Different future caching strategies mitigate this behavior differently.

tomdale:
@rwjblue Given our most recent conversation, should we close this PR and wait for your config changes to Ember CLI?

rwjblue:
Yes, makes sense to me.

PR #85, opened 2016-01-17
Add fastboot-config with whitelisted modules

tomdale:
This commit adds generation of a `fastboot-config.json` to the `fastboot-dist`. For now, this only includes one field which is a list of npm packages that are needed when the app is running in FastBoot.

For example, a `fastboot-config.json` may look like this:

```json
{
  "moduleWhitelist": ["najax", "node-fetch"]
}
```

The list of whitelisted modules is derived from the app's installed addons. Addons can add a `fastBootDependencies` key to their `package.json`'s `ember-addon` field:

```json
{
  "name": "ember-network",
  "version": "0.1.0",
  "ember-addon": {
    "fastBootDependencies": ["node-fetch"]
  }
}
```

Each addon's list of dependencies will be aggregated and placed into a single list of whitelisted modules.

### DIFF
--- a/lib/utilities/fastboot-app-module.js
+++ b/lib/utilities/fastboot-app-module.js
@@ -1,0 +1,25 @@
+// Expose the an factory for the creating the `Application` object
+// with the proper config at a known path, so that the server does
+// not have to disover the app's module prefix ("my-app").
+//
+// The module defined here is prefixed with a `~` to make it less
+// likely to collide with user code, since it is not possible to
+// define a module with a name like this in the file system.
+function fastbootAppModule(prefix) {
+  return [
+    "",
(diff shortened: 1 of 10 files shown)


lib/utilities/fastboot-app-module.js
```diff
@@ -0,0 +1,25 @@
+// Expose the an factory for the creating the `Application` object
```
rwjblue: The an factory?

tomdale: I was just refactoring the existing code. :) I plan to do a big cleanup once this is merged, including extracting the testing stuff.

rwjblue: Yeah, after commenting I saw that this was the same verbiage we already have.

PR #88, opened 2016-01-19
Add Ember 2.3 requirement to README

michaelrkn:
Since FastBoot can now be used with Ember 2.3 stable, this is the best path for most users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,22 +13,13 @@ your production app and have FastBoot work.
 
 ## Installation
 
+FastBook requires Ember 2.3 or higher.
+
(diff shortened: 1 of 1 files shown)


tomdale:
Thanks!

cs3b:
I would not be so fast - I didn't manage to run FastBoot with 2.3 (after upgrading to  it start working smoothly).

Full story described here: https://github.com/tildeio/ember-cli-fastboot/issues/87 

And here is commit the fix the issue in ember: https://github.com/emberjs/ember.js/pull/12829

PR #91, opened 2016-01-19
Extract test helpers

tomdale:


tomdale:
Seems like it is failing on Travis, will have to investigate.

Issue #92, opened 2016-01-20
Allow to set hostname that server will bind to.


cs3b:
When deploying fastboot to staging environment (on Heroku), I've encounter problem that heroku expects you to bind port to 0.0.0.0 (IPv4). I've introduced another option to command line in order to get there. 

https://github.com/cs3b/ember-cli-fastboot/commit/fa73a46f5e8301f52c0c0002adc7ff8abf71b7ff

It somehow relates to: https://github.com/tildeio/ember-cli-fastboot/issues/58 and https://github.com/tildeio/ember-cli-fastboot/issues/46

PR #93, opened 2016-01-20
Allow to define host as parameter, and set the default as [::]

cs3b:
So know you can `ember fastboot --host 0.0.0.0`

solve #92


tomdale:
@cs3b Awesome, thank you

PR #95, opened 2016-01-21
Replace ajax{Success,Error} with an override of `adapter._ajaxRequest`

danmcclain:
Per @tchak's suggestion, moving the jQuery request in the Ember Data
adapter [here][ed-pr], the code to replace `jQuery.ajax` with `najax`
becomes trivial, and replaces all code duplication.

There was an update to najax that is needed for Ember data to be able to
parse the headers. That code is available on the master branch of najax,
but has not been released.

This also depends on the latest beta branch of Ember Data, and will be
compatible with ED 2.4

[ed-pr]: emberjs/data#4113

### DIFF
--- a/app/initializers/server/ajax.js
+++ b/app/initializers/server/ajax.js
@@ -1,21 +1,7 @@
-/*globals najax, Ember*/
-var nodeAjax = function(url, type, options) {
-  var adapter = this;
+/* globals najax */
 
-  return new Ember.RSVP.Promise(function(resolve, reject) {
-    var hash = adapter.ajaxOptions(url, type, options);
-
-    hash.success = function(json, textStatus, jqXHR) {
-      json = adapter.ajaxSuccess(jqXHR, json);
-      Ember.run(null, resolve, json);
-    };
-
-    hash.error = function(jqXHR, textStatus, errorThrown) {
-      Ember.run(null, reject, adapter.ajaxError(jqXHR, jqXHR.responseText, errorThrown));
-    };
-
-    najax(hash);
-  }, 'DS: RESTAdapter#ajax ' + type + ' to ' + url);
+var nodeAjax = function(options) {
+  najax(options);
 };
 
 export default {
@@ -23,6 +9,6 @@ export default {
 
   initialize: function(application) {
     application.register('ajax:node', nodeAjax, { instantiate: false });
-    application.inject('adapter', 'ajax', 'ajax:node');
+    application.inject('adapter', '_ajaxRequest', 'ajax:node');
   }
 };


fivetanley:
These error classes are not private, we just don't have a file to export them I guess. But the documentation shows them as public right now. You should be able to yoink them off DS.

danmcclain:
@fivetanley you yoink them off DS would still make ember data a requirement of fastboot, not sure if that is desired, attempting to avoid it

tchak:
I d'ont understand this PR. Why ember-data is leaking in to fastboot network code ? The whole refactor of network code in ember-data was made to abstract network layer. I something need to be fixed in ember-data to avoid this - let's talk and find a way to fix it.

tchak:
Ok, if I understand correctly the problem is that we need to call `najax` instead of `jQuery.ajax`. I have two solutions to this:
- define `jQuery.ajax` in fastboot.
- expose `ajaxFn` property on `RestAdapter`.

tchak:
`application.inject('adapter', 'ajax', 'ajax:node');` shouldn't it be `application.inject('rest-adapter', 'ajax', 'ajax:node');`

danmcclain:
I have a PR on Ember Data that ducktypes the adapter errors, I have yet to update this PR to reflect those changes, long story short: the ember data errors will be removed from this PR @tchak

danmcclain:
@tchak Let me spike out the suggestion of using `ajaxFn` on Ember Data, I'll have something in a little bit

danmcclain:
@tchak Also, the injection has to be on `adapter` not `rest-adapter`, I'm not entirely sure why though. A trivial spike of the `ajaxFn` makes this code obsolete, let me PR ED and fastboot,

tchak:
@danmcclain cool :+1:

danmcclain:
Much cleaner approach thanks to @tchak, will squash PRs once the this PR closes: https://github.com/emberjs/data/pull/4113

danmcclain:
@tomdale this has been updated and depends on ember-data#beta and najax#master

tomdale:
@danmcclain I wonder if there's a way we could do this without breaking folks who aren't on ED beta channel?

danmcclain:
@tomdale You need to be on Ember-Data < 2.0.0 for fastboot to work the way it is now, as `ajaxSuccess` and `ajaxError` were removed then

2.0.0: https://github.com/emberjs/data/blob/v2.0.0/packages/ember-data/lib/adapters/rest-adapter.js#L745-L755

1.13.15: https://github.com/emberjs/data/blob/v1.13.15/packages/ember-data/lib/adapters/rest-adapter.js#L809-L813

tomdale:
@danmcclain Got it

fivetanley:
I am confused as to why ember data is required at all for fastboot to work?

tomdale:
It is not

fivetanley:
Ah it is only compatible with certain versions, I see.

tchak:
@tomdale I hope we can switch `ember-data` to `ember-network` soon so we would not need to go through all this hoops...

PR #110, opened 2016-02-03
Tests for environment-specific initializers

tomdale:
Tests for filtering out initializers and instance initializers. The rules are:

1. Initializers in `app/initializers/browser/` are not included in FastBoot builds.
2. Initializers in `app/initializers/fastboot/` are not included in browser builds
3. Initializers in `app/initializers/` are included in both.

(Similar rules apply to instance initializers.)

### DIFF
--- a/lib/addon-modes/browser.js
+++ b/lib/addon-modes/browser.js
@@ -1,10 +1,11 @@
 var filterInitializers = require('../broccoli/filter-initializers');
+
 module.exports = {
   /**
    * Filters out initializers and instance initializers that should only run in
    * FastBoot mode.
    */
-  treeForApp: function(tree) {
-    return filterInitializers(tree, 'server');
+  preconcatTree: function(tree) {
+    return filterInitializers(tree, 'fastboot', this.app.name);
   }
 };
--- a/lib/addon-modes/fastboot.js
+++ b/lib/addon-modes/fastboot.js
@@ -1,4 +1,3 @@
-var path       = require('path');
 var mergeTrees = require('broccoli-merge-trees');
 var Funnel     = require('broccoli-funnel');
 
@@ -36,14 +35,12 @@ module.exports = {
     return CONFIG;
   },
 
+  /**
+   * Filters out initializers and instance initializers that should only run in
+   * browser mode.
+   */
   preconcatTree: function(tree) {
-    return new Funnel(tree, {
-      annotation: 'Funnel: Remove browser-only initializers',
-      exclude: [
-        path.join(this.app.name, 'initializers/browser/*'),
-        path.join(this.app.name, 'instance-initializers/browser/*')
-      ]
-    });
+    return filterInitializers(tree, 'browser', this.app.name);
   },
 
   /**
@@ -66,14 +63,6 @@ module.exports = {
     }
 
     return tree;
-  },
-
-  /**
-   * Filters out initializers and instance initializers that should only run in
-   * browser mode.
-   */
-  treeForApp: function(tree) {
-    return filterInitializers(tree, 'browser');
   },
 
   /**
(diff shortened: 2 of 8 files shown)


chancancode:
:eyes: seems good!

rwjblue:
What does this mean for addons that want to be consumable for fastboot?

For example, if addons want to make their addon work in fastboot by providing alternative `app/initializers/server/foo.js` and `app/initializers/browser/foo.js` files, they would have to do this filtering themselves in their own `app` trees (so that it works properly for apps that do not include `ember-cli-fastboot`).

rwjblue:
My suggestion is to make this `filter-initializers` a separate npm package that addons can use to mix in and filter their own app trees automatically.

lib/addon-modes/browser.js
```diff
@@ -1,10 +1,11 @@
 var filterInitializers = require('../broccoli/filter-initializers');
+
 module.exports = {
   /**
    * Filters out initializers and instance initializers that should only run in
    * FastBoot mode.
    */
-  treeForApp: function(tree) {
-    return filterInitializers(tree, 'server');
+  preconcatTree: function(tree) {
```
rwjblue: Why are we creating our own hook here? In general, I would prefer to stick with the normal hooks like `treeForApp`.  If we need to make extra hooks upstream because something is not possible otherwise let's do it, but I'm not seeing exactly why we would want to do this before `concatFiles` vs `treeForApp`...

tomdale:
@rwjblue I think there might be some confusion about what this does.

1. Previously, `ember-fastboot-cli` filtered out its own initializers and instance initializers based on what the target runtime was (browser or FastBoot).
2. Because this pattern is generally useful, I have expanded it to apply to the _final_ app tree, so all addons (and the app itself!) can take advantage of it. If you want initializers that only run in the browser, put them in `app/initializers/browser/foo.js`.
3. There is no Ember CLI addon hook for munging the `app/` tree after the app and the addons have been merged. This is why I added a `preconcatTree` hook. I plan to upstream this to Ember CLI as I agree monkeypatching is not a long-term strategy.
4. `treeForApp` is the tree for the _addons_ `app/` directory only; it doesn't let us change how other apps and addons work.

Because this is such a common need, I don't think it's fair to make app developers learn about Broccoli trees, how to change theirs, and know that an npm package for this exists, just to get environment-dependent initializers. I also want to make the names conventional; if we require addon others to write their own Broccoli filters, we'll have a hundred different naming schemes.

rwjblue:
@tomdale - As discussed in slack I was not about the browser/fastboot initializer concepts.  Let me paint a picture:

* We want addon authors to be able to use `app/initializers/{fastboot,browser}`.
* Addons are sometimes used in fastboot, but more often used in non-fastboot apps.
* When an addon that includes `app/initializers/fastboot/foo.js` is used in an application that is not using `ember-cli-fastboot` the fastboot initializer will run in the browser.  This is bad. :smiling_imp: 

So, my suggestion from https://github.com/tildeio/ember-cli-fastboot/pull/110#issuecomment-179545760 still needs to happen so that addon authors can always ensure their addon works (ember-cli-fastboot or not).

ronco:
I just tested this out in browser with a fastboot-less app and I'm seeing that exact behavior (both initializers running). I could potentially work around it by putting a gaurd around the fastboot initializer to only run if Fastboot is present, but that feels redundant.

tomdale:
@rwjblue Yes I totally agree.
@ronco I think @rwjblue has the right idea that we should release a small Broccoli filter addon authors can use and include to filter their `appTree`. I can probably do it sometime this week but probably not before my talk in Romania as I'm rushing hard to get documentation and my presentation ready.

ronco:
I can take a shot at it tonight. That branch on ember-cli-head should be functional in fastboot environments until then.

tomdale:
@ronco Did that get extracted?

ronco:
@tomdale I started working on it last week but haven't had a chance to finish it up.  I do have a couple questions I'd like to hit you up on that might speed it up when I get back to it tonight.  I'll hit you up on slack shortly.

PR #113, opened 2016-02-10
Remove unused functions

jpadilla:
These functions are leftover from f87c8eb3b6cf3fa7588e9eb45a18fd20e7f61fab and no longer needed.

### DIFF
--- a/lib/commands/fastboot.js
+++ b/lib/commands/fastboot.js
@@ -1,5 +1,3 @@
-var path = require('path');
-var fs = require('fs');
 var RSVP = require('rsvp');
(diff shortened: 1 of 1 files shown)


rwjblue:
Can you also move `glob` to be a `devDependency` now?

jpadilla:
@rwjblue good call. Also noticed that `chalk`, `debug`, and `silent-error` don't seem to be used. Should I remove those as well?

tomdale:
@jpadilla That would be great. Are you sure `debug` isn't used? I think everything that used `chalk` got moved to `ember-fastboot-server`.

PR #120, opened 2016-02-15
[WIP] Add cookies service

marcoow:
The cookies service wraps cookie access via either express.js' request/response (if in fastboot mode) or `document.cookie` if in browser mode. This is a prerequisite for implementing a shared session between browser and fastboot mode.

Would be great to get some feedback on whether this approach is even seen as a good idea or everybody had sth. else in mind.


tomdale:
Closing in favor of #121.

PR #123, opened 2016-02-25
Add headers to FastBoot service

danmcclain:


danmcclain:
@tomdale will need a new release of ember-fastboot-server before this passes

Issue #126, opened 2016-02-27
Cannot access window.location


paulelliott:
I need to inspect the current subdomain and get the full URL for use with Open Graph tags but `window.location` is not available when running in fastboot.

tomdale:
I'm not sure how this would work. In the browser you can derive `window.location` because it's the hostname of the server that gave you the assets, a.k.a. the origin. But on a node server probably running behind a load balancer, I don't think there's any way we could reliably guess this. It seems like a better bet would be to include the domain in your environment/configuration set up and use that.

paulelliott:
I was thinking you could either:

1) Wrap the referrer header with a mock location object.

2) Just document that this doesn't work in fastboot. Maybe have an empty dummy object with a configuration like you suggest so that it doesn't blow up.

paulelliott:
For a little background on why this is important in my case, this ember app is publicly accessible and runs across multiple subdomains. Customers can choose a subdomain and the API server uses that as an account key to know what data to return. Since it is one app running multiple accounts, I can't just set that in the environment.

tomdale:
@paulelliott We've tried to avoid recreating a synthetic browser environment (à la mimicking `window.location`) but I am definitely in favor of exposing this on the `fastboot` service.

paulelliott:
That fastboot service already has `headers` and `cookies` in it so it seems to be shaping up to be a representation of the request itself. I wouldn't want to just add the raw request to it because that couples fastboot to the underlying implementation.

If that is the direction then I would suggest just adding `currentUrl` as an attribute of the service directly.

Otherwise I would nest a third object alongside `headers` and `cookies` because that opens the door for more functionality down the road. I wouldn't want to call it `request` because it implies that it is the actual request but it isn't. 

What do you think of one of these contracts?
```
  this.get('fastboot.currentUrl')

  OR

  this.get('fastboot.meta.url')
  this.get('fastboot.requestData.urll')
  this.get('fastboot.requestMeta.url')
  ...or some other container name
```

tomdale:
@paulelliott In #128, @danmcclain has exposed a `host` property. That would solve your use case, right?

One thing I'm nervous about are the security implications of this. Here's why:

In the browser, you are de facto running on an untrusted, remote client. Because you have designed for a 100% untrusted environment, the user forging things like URLs doesn't matter. For example, if you're using `window.location` to construct a new URL, it doesn't matter if the user tampers with it because they could just cURL that URL anyway.

[...]

PR #128, opened 2016-02-27
Allow ember-data to work when `host` is not defined

danmcclain:
This uses the headers/request fron express to figure out where to make
the najax request to.

This is in lieu of a proper jQuery AJAX transport

### DIFF
--- a/app/initializers/fastboot/ajax.js
+++ b/app/initializers/fastboot/ajax.js
@@ -1,6 +1,19 @@
 /* globals najax */
+import Ember from 'ember';
+
+const { get } = Ember;
 
 var nodeAjax = function(options) {
+  let httpRegex = /^https?:\/\//
+
+  if (!httpRegex.test(options.url)) {
+    try {
+      options.url = get(this, 'fastboot.host') + options.url
+    } catch (fbError) {
+      throw new Error('You are using Ember Data with no host defined in your adapter. This will attempt to use the host of the FastBoot request, which is not configured for the current host of this request. Please set the hostWhitelist property for in your environment.js. FastBoot Error: ' + fbError.message);
+    }
+  }
+
   najax(options);
 };
 
@@ -10,5 +23,6 @@ export default {
   initialize: function(application) {
     application.register('ajax:node', nodeAjax, { instantiate: false });
     application.inject('adapter', '_ajaxRequest', 'ajax:node');
+    application.inject('adapter', 'fastboot', 'service:fastboot');
   }
 };


app/initializers/fastboot/ajax.js
```diff
@@ -1,6 +1,17 @@
 /* globals najax */
+import Ember from 'ember';
+
+const { get } = Ember;
 
 var nodeAjax = function(options) {
+  let httpRegex = /^https?:\/\//
+
+  if (!httpRegex.test(options.url)) {
+    let host = `${get(this, 'fastboot._fastbootInfo.reques.protocol')}://${get(this, 'fastboot.headers.host')}`;
```
tomdale: I have a doubt

tomdale: I think we need tests for this. :trollface:

tomdale:
This is also needed by #126; we should think about exposing this info as public API to support @paulelliott's use case as well.

danmcclain:
@tomdale Added `host` to the service, PR to `ember-fastboot-server`: https://github.com/ember-fastboot/ember-fastboot-server/pull/25

paulelliott:
This would solve my issues, for sure. Any ETA on when we could see this merged?

danmcclain:
@paulelliott #136 replaces this PR in terms of introducing the host API. Once #136 is merged, I'm going to update this issue to use the new host API

danmcclain:
@tomdale this is rebased, and should be good to merge

tomdale:
@danmcclain This should probably have a better error message if there isn't a host whitelisted, right? People randomly using Ember Data will hit it and the error message about accessing `fastboot.host` will probably be very confusing.

danmcclain:
@tomdale Added a long winded, informative error message :smile:

app/initializers/fastboot/ajax.js
```diff
@@ -1,6 +1,19 @@
...
+const { get } = Ember;
 
 var nodeAjax = function(options) {
+  let httpRegex = /^https?:\/\//
+
+  if (!httpRegex.test(options.url)) {
+    try {
+      options.url = get(this, 'fastboot.host') + options.url
+    } catch (fbError) {
+      throw new Error('You are using Ember Data with no host defined in your adapter. This will attempt to use the host of the FastBoot request, which is not configured for the current host of this request. Please set the hostWhitelist property for in your environement.js. FastBoot Error: ' + fbError.message);
```
acorncom: Right at the very end of this long error message `environment` is mis-spelled :-)

danmcclain: Thanks!

danmcclain:
@tomdale Ping

tomdale:
Mérgèd

PR #137, opened 2016-03-14
Update README.md

milworm:
fixed typo:
"either nor" -> "neither nor"

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ consider deploying this to production.
 
 Since `didInsertElement` hooks are designed to let your component
 directly manipulate the DOM, and that doesn't make sense on the server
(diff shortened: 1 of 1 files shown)


tomdale:
Thanks for the PR. I think the correct fix is to change "nor" to "or." In English, the proposed change introduces a double negative which would change the meaning of the sentence.

The correct version should be either:

> where there is no DOM, we do not invoke either the `didInsertElement` or `willInsertElement` hooks.

or:

> where there is no DOM, we invoke neither `didInsertElement` nor `willInsertElement` hooks.

milworm:
Sorry, my bad. Forgot to check what was written before "either or".

PR #139, opened 2016-03-16
Produce both browser and FastBoot build

tomdale:
This also moves asset rewriting into the build step


tests/acceptance/asset-rewriting-test.js
```diff
@@ -0,0 +1,28 @@
...
+    app = new AddonTestApp();
+
+    return app.create('dummy');
+  });
+
+  it("rewrites index.html to point to browser build", function() {
+    return app.runEmberCommand('fastboot:build', '--environment=production')
+      .then(function() {
+        var appPath = glob.sync(app.filePath('fastboot-dist/browser/assets/dummy-*.js'))[0];
+        var matches = appPath.match(/fastboot-dist\/browser\/assets\/dummy-(.*).js/);
```
danmcclain: Does this account for CDN assets?

topaxi:
Tests are passing with #140 :)

tomdale:
@topaxi LOL, just pushed a rebased version right as you commented. ;) Hopefully this does the trick…

index.js
```diff
@@ -35,36 +37,80 @@ module.exports = {
   included: function(app) {
     patchEmberApp(app);
 
-    var mode;
-
-    if (isFastBoot()) {
-      process.env[APP_NAME_KEY] = app.name;
-      app.options.autoRun = false;
+    // We serve the index.html from fastboot-dist, so this has to apply to both builds
+    app.options.storeConfigInMeta = false;
```
ef4: Why does fastboot need `storeConfigInMeta = false`?

tomdale: @ef4 Unsure. @rwjblue/@chancancode, can we remove this?

stefanpenner:
I have brought this up before, but why are multiple builds required. it seems like fast-boot bootstrapping code could be made to handle all the differences either by wrapping or utilizing extra parametrization. This will likely require improving the bootstraping code for normal apps as well.

Having this branching has downsides:

* increasing the cost of builds
* increasing the chances some part of the stack doesn't compose well

index.js
danmcclain: Is it because the server doesn't read th index.html, so when the app.js is evaluated for the server, the config will be missin?

rwjblue: Without this ember-cli creates a module for the app which uses a jQuery selector to find a meta tag in the `index.html` (which the fastboot started app doesn't have access to).  We need to change the way ember-cli generates the config and emits the index.html. I have generally been working on it, but have fallen behind a bit :crying_cat_face:.

tomdale:
Gonna merge this, but I agree with @stefanpenner that, with the new approach, having a shared `app.js` is within reach. I will work on that next, though we will still want to produce a per-environment `vendor.js`.

stefanpenner:
> produce a per-environment vendor.js.

sounds good

PR #142, opened 2016-03-25
Add isFastboot property to service

davidpett:
A convenience property to allow users to check if it is in fastboot mode or not

### DIFF
--- a/app/services/fastboot.js
+++ b/app/services/fastboot.js
@@ -1,3 +1,4 @@
+/* global FastBoot */
 import Ember from "ember";
 
 let alias = Ember.computed.alias;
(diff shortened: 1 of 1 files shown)


tomdale:
Should be `isFastBoot`

rwjblue:
> Should be isFastBoot

We should have a "Annoying Capitalization Details" section in the README. :smiling_imp:

davidpett:
Changed it to be proper, good sir.

tomdale:
@davidpett So looking at this more closely, I think better check is not `window` or `document` but to see if the `FastBoot` global exists. Sorry I didn't see that earlier!

davidpett:
Ahh, didn't realize that was a thing. Not sure how to create a test for this, but I did test it in my app and it works as advertised. EVERYBODY GETS A FASTBOOT!

PR #144, opened 2016-03-26
[bugfix] work around duplicated babel plugin options

fivetanley:
When another addon, like Ember Data, uses a babel plugin to change code
at build time, babel (provided by ember-cli-babel) throws an exception
that the plugin cannot be included twice because it has the same name.
This is happening because copying `plugins` of the original app's
`options` includes addon's babel plugins. Passing those options to `new
EmberApp()` causes the plugins to get loaded twice by `EmberApp`'s
loading logic.

I don't know what the right solution is. Maybe ember-cli-babel should
de-dupe options. Maybe we need some way to get at the `options` hash
that was present before the `EmberApp`'s initializers were run.

### DIFF
--- a/lib/broccoli/fastboot-build.js
+++ b/lib/broccoli/fastboot-build.js
@@ -57,6 +57,16 @@ FastBootBuild.prototype.appOptions = function() {
   options.autoRun = false;
   options.project = this.buildFastBootProject();
 
+  // If plugins get included twice, babel complains about duplicate
+  // plugins with the same name and will throw an error that
+  // causes the build to fail.
+  // We can remove them here and they should hopefully be picked up
+  // by `new EmberApp()`. I don't know if this excludes plugins that
+  // are included by putting babel options in `ember-cli-build.js`.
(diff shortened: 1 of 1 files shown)


stefanpenner:
@fivetanley i would consider a PR to ember-cli-babel or broccoli-babel-transpiler that uniq's these and warns or something

fivetanley:
@stefanpenner we could do that but that leaves users with a warning they can't fix. i think post ember-conf we should try and look at some areas where we can re-run the `EmberApp` more easily.

fivetanley:
Wait, shouldn't we be able to `require` ember-cli-build.js instead of this? If it returns a `new EmberApp`, that seems good.

stefanpenner:
> Wait, shouldn't we be able to require ember-cli-build.js instead of this? If it returns a new EmberApp, that seems good.

glimmer does this

fivetanley:
I'm closing this in favor of that approach. Thanks stef!!

Issue #141, opened 2016-03-24
Asset in template is not fingerprinted in production


topaxi:
Using current ember-cli-fastboot master.

If I have a template with an image, it refers to the unfingerprinted asset.

Input:
```hbs
<img src="/assets/foo.png">
```

Output:
```hbs
<img src="/assets/foo.png">
```

Expected:
```hbs
<img src="/assets/foo-SomeHash.png">
```

See: [topaxi.ch](https://topaxi.ch/) The `puzzle.png` image
Or the repository: https://github.com/topaxi/topaxi.ch

Once the the page loads, the browser build correctly uses the fingerprinted asset.

thejase:
Experiencing this same issue.  Build strategy:

```
ember build --environment=production # builds to /dist
ember fastboot --environment production --port=1234 --serve-assets
```

jasonmit:
https://github.com/tildeio/ember-cli-fastboot/blob/master/lib/broccoli/fastboot-build.js#L35

It's only asset revving over the javascript tree, so it doesn't take into account anything in the public tree which is why it misses `foo.png`.  I worked around this by including the full `emberApp.proto.toArray` `sourceTree`.  That's when I also uncovered that everything in the root of `/public` (robots.txt, crossdomain.xml, etc.) isn't served up by the fastboot server.

Unsure how to resolve this without resolving the new bug I stumbled on.

thejase:
The workaround I found was just to build using a non-production mode.  Unfortunately, this removes build signatures from all assets, so we have to resolve back to browser caching, which is at least one refresh behind, no matter your caching strategy.

ebryn:
Removing `broccoli-asset-rev` from your package.json is a quick and easy workaround for now.

ballPointPenguin:
not sure if this is the same issue, but I see 404s on my fingerprinted css and js assets despite the fact that the ember app is running correctly

PR #145, opened 2016-03-26
consume as a broccoli tree instead of re-creating EmberApp

fivetanley:
This should fix an issue where babel plugins get re-registered twice.
There's a myriad of situations where copying over the options from an
existing application could result in misconfigurd or duplicate builds.

Instead of ripping `options` off EmberApp, we use the function exported
by the Ember CLI application's `ember-cli-build.js` file to return a
broccli tree. This allows the app to use any broccoli plugins (without
creating an addon) and also re-runs any initialization steps we might
have missed by creating the `EmberApp` ourselves. This also eliminates
reliance on the private `javascript()` method on `EmberApp`.

### DIFF
--- a/lib/broccoli/fastboot-config.js
+++ b/lib/broccoli/fastboot-config.js
@@ -3,6 +3,7 @@ var fmt    = require('util').format;
 var uniq   = require('lodash.uniq');
 var path   = require('path');
 var Plugin = require('broccoli-plugin');
+var defaults = require('lodash.defaults');
 
 function FastBootConfig(inputNode, options) {
   Plugin.call(this, [inputNode], {
@@ -82,7 +83,6 @@ FastBootConfig.prototype.readAssetManifest = function() {
 
   try {
     var assetMap = JSON.parse(fs.readFileSync(assetMapPath));
-    fs.unlinkSync(assetMapPath);
     return assetMap;
   } catch (e) {
     // No asset map was found, proceed as usual
@@ -103,13 +103,19 @@ FastBootConfig.prototype.buildManifest = function() {
   var rewrittenAssets = this.readAssetManifest();
 
   if (rewrittenAssets) {
+    var assets = {};
+
+    // Because the assetMap was written before the files in the
+    // `asset` folder were moved to the `fastboot` folder,
+    // we have to rewrite them here.
+    for (var key in rewrittenAssets.assets) {
+      var rewrittenKey = assetToFastboot(key);
+      assets[rewrittenKey] = assetToFastboot(rewrittenAssets.assets[key]);
+    }
+
     ['appFile', 'vendorFile'].forEach(function(file) {
-      // The file generated for the assetMap.json will point
-      // to assets. After the build, we move it to the
-      // `fastboot` directory so we need to update it accordingly.
-      var asset = rewrittenAssets.assets[manifest[file]];
-      var fileName = path.basename(asset);
-      manifest[file] = path.join('fastboot', fileName);
+      // Update package.json with the fingerprinted file.
+      manifest[file] = assets[manifest[file]];
     });
   }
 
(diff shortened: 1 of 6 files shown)


fivetanley:
I'm testing this against the ember-api-docs now to make sure extra files aren't included in `fastboot` folder, I'll post back shortly

fivetanley:
One downside to this approach is that we end up building css/etc twice when we only want to build JS twice. I think we can configure this with options:

https://github.com/ember-cli/ember-cli/blob/65f6ffafc9f53f1ced3ea9563ef2b27c40763763/lib/broccoli/ember-app.js#L220

fivetanley:
Some of the project directives like the `appFile` field are missing in package.json. I'm looking into that.

jasonmit:
I was trying this branch out because it touched some similar logic I was working on to fix #141.  However, it looks like production builds are failing I suspect due to the changes around asset map.

```sh
ember fastboot --environment production --port=1234       ⏎ ✖ ✚ ✭
version: 2.4.1
Built project successfully. Stored in "fastboot-dist".
Installing FastBoot npm dependencies
Error: ENOENT: no such file or directory, open 'fastboot-dist/fastboot/asset-rev-example.js'
    at Error (native)
    at Object.fs.openSync (fs.js:584:18)
    at Object.fs.readFileSync (fs.js:431:33)
    at new EmberApp (/home/elek/projects/ember-cli-fastboot/node_modules/ember-fastboot-server/lib/ember-app.js:39:20)
    at new FastBootServer (/home/elek/projects/ember-cli-fastboot/node_modules/ember-fastboot-server/lib/server.js:33:14)
    at /home/elek/projects/ember-cli-fastboot/lib/commands/fastboot.js:32:22
    at lib$rsvp$$internal$$tryCatch (/home/elek/projects/ember-cli-fastboot/node_modules/rsvp/dist/rsvp.js:1036:16)
    at lib$rsvp$$internal$$invokeCallback (/home/elek/projects/ember-cli-fastboot/node_modules/rsvp/dist/rsvp.js:1048:17)
    at lib$rsvp$$internal$$publish (/home/elek/projects/ember-cli-fastboot/node_modules/rsvp/dist/rsvp.js:1019:11)
    at lib$rsvp$asap$$flush (/home/elek/projects/ember-cli-fastboot/node_modules/rsvp/dist/rsvp.js:1198:9)
    at doNTCallback0 (node.js:430:9)
    at process._tickCallback (node.js:359:13)
```

fivetanley:
@jasonmit I figured out the source of that (which was also the source of my test failure). Due to this change, you'll need to enable `generateAssetMap: true` in the `fingerprint` options passed to `new EmberApp` in your Brocfile.

fivetanley:
I need to figure out a way to cancel the build if this isn't present...

fivetanley:
An example: https://github.com/fivetanley/ember-cli-fastboot/blob/require-ember-app/tests/fixtures/module-whitelist/ember-cli-build.js#L5

lib/broccoli/fastboot-config.js
```diff
@@ -82,7 +83,6 @@ FastBootConfig.prototype.readAssetManifest = function() {
 
   try {
     var assetMap = JSON.parse(fs.readFileSync(assetMapPath));
-    fs.unlinkSync(assetMapPath);
```
fivetanley: this change is no longer needed, i was trying to debug something.

danmcclain:
I can verify that this works, I've been using this PR for deploying dockyard.com

sepowitz:
hi all - any reason this hasn't been merged yet?  happy to help!

tomdale:
@sepowitz Just my own laziness. :) I am reviewing it now and hope to have it merged soon.

hhff:
:+1:

sepowitz:
@tomdale right on!

tomdale:
@fivetanley So this looks great. I started rebasing this onto master, and I think there may be a regression in this refactor in that it looks like output paths are not respected. I pushed the rebased version to the `fivetanley-require-ember-app` branch; can you take a look? It has one failing test, which is that the manifest is using the `outputPaths` but it is not finding anything there.

hhff:
ya I noticed that @fivetanley's branch is against 0.6.0, but we're at 0.6.2

PR #146, opened 2016-03-27
Automatically adds /fastboot-dist to project .gitignore

jasonmit:
Doesn't need much explaining, just adds `/fastboot-dist` to `.gitignore` when ember-cli-fastboot is installed.

### DIFF
--- a/blueprints/ember-cli-fastboot/index.js
+++ b/blueprints/ember-cli-fastboot/index.js
@@ -1,0 +1,17 @@
+/*jshint node:true*/
+
+'use strict';
+
+module.exports = {
+  normalizeEntityName: function() {
+    // this prevents an error when the entityName is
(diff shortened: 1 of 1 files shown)


danmcclain:
I don't think the fastboot-dist is created anymore

jasonmit:
@danmcclain is that the plan or on a private branch?  I'm seeing fastboot-dist based on a build from master.

danmcclain:
I thought that https://github.com/tildeio/ember-cli-fastboot/pull/139 accomplished this

@tomdale can you confirm/deny?

tomdale:
It's still created if you use `ember fastboot:build`, which is deprecated. It's currently also created when you run `ember fastboot`, but that was just an oversight and I'll change it. I'd rather have everyone start using `dist` than add it to the gitignore. :)

PR #169, opened 2016-04-13
Implement a function to defer rendering via promise chaining

danmcclain:
Service side of
https://github.com/ember-fastboot/ember-fastboot-server/pull/41

### DIFF
--- a/app/services/fastboot.js
+++ b/app/services/fastboot.js
@@ -48,5 +48,10 @@ export default Ember.Service.extend({
 
   isFastBoot: computed(function() {
     return typeof FastBoot !== 'undefined';
-  })
+  }),
+
+  deferRendering(promise) {
+    Ember.assert('deferRendering requires a promise or thennable object', promise.then);
+    this._fastbootInfo.deferRendering(promise);
+  }
 });
--- a/tests/acceptance/async-content-test.js
+++ b/tests/acceptance/async-content-test.js
@@ -1,0 +1,38 @@
+var chai = require('chai');
+var expect = chai.expect;
+var RSVP = require('rsvp');
+var AddonTestApp = require('ember-cli-addon-tests').AddonTestApp;
+var request = require('request');
+var get = RSVP.denodeify(request);
+
+describe('async content via deferred content', function() {
+  this.timeout(300000);
+
+  var app;
+
+  before(function() {
+
+    app = new AddonTestApp();
+
+    return app.create('async-content')
+      .then(function() {
+        return app.startServer({
+          command: 'fastboot',
+          additionalArguments: ['--serve-assets']
+        });
+      });
+  });
+
+  after(function() {
+    return app.stopServer();
+  });
+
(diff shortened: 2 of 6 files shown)


tests/fixtures/async-content/app/components/async-content.js
```diff
@@ -0,0 +1,20 @@
...
+    this._super(...arguments);
+
+    let deferred = Ember.RSVP.defer();
+
+    Ember.run.later(() => {
+      if (!this.get('isDestroyed')) {
+        this.set('setLater', 'foo');
+      }
+      deferred.resolve()
+    }, 100);
```
danmcclain: This is enough of a delay to escape the initial `visit` call

danmcclain:
@tomdale @stefanpenner Updated per comments on #180

app/services/fastboot.js
```diff
@@ -46,5 +46,8 @@ export default Ember.Service.extend({
 
   isFastBoot: computed(function() {
     return typeof FastBoot !== 'undefined';
-  })
+  }),
+  deferRendering(promise) {
```
stefanpenner: don't we more typically have hooks, the when invoked do.

new Promise(resolve => resolve(cb());

Passing promises around like this doesn't really align with other hooks, like model routes hooks, or adapter hooks.

tests/acceptance/async-content-test.js
```diff
@@ -0,0 +1,39 @@
...
+
+  var app;
+
+  before(function() {
+
+    app = new AddonTestApp();
+
+    return app.create('async-content')
+      .then(function() {
+        debugger;
```
stefanpenner: ^

tests/fixtures/async-content/app/components/async-content.js
```diff
@@ -0,0 +1,37 @@
...
+  },
+
+  willDestroy() {
+    let runLater = this.get('runLater');
+    let deferred = this.get('deferred');
+
+    if (runLater) {
+      Ember.run.cancel(runLater);
+      deferred.promise.catch(function() {
+        console.log('swallow reject');
```
stefanpenner: we should let it throw, so it hits `Ember.onerror`

tests/fixtures/async-content/app/components/async-content.js
```diff
@@ -0,0 +1,37 @@
...
+      this.set('deferred', null);
+    }, 100);
+
+    this.runLater =  runLater;
+    this.deferred = deferred;
+
+    this.get('fastboot').deferRendering(deferred.promise);
+  },
+
+  willDestroy() {
```
stefanpenner: `destroy`

app/services/fastboot.js
danmcclain: If we were to add a hook, where would we introduce it?

This is to handle the case that you have asynchrony outside of the normal hooks. as such, you may be doing this work inside a component.

Or are you suggesting that we make a new promise from the promise being passed into `deferRendering`

PR #175, opened 2016-04-19
Add list of hooks called in FastBoot

danmcclain:


rwjblue:
IMHO, this should also be made clear in the Ember API docs.

rwjblue:
Some updates around [here](https://github.com/emberjs/ember.js/blob/master/packages/ember-views/lib/components/component.js#L483-L593) would do it...

PR #177, opened 2016-04-20
Add Ember.onerror hook for fastboot apps to prevent node from crashing

kratiahuja:
This PR addresses the issue of node crashing when there is an exception running your app in fastboot environments.

When there are errors in the run loop due to accessing DOM API or any other errors, the node process crashes. This initializer catches and logs such errors and prevents the app from crashing.

I have manually tested it with a dummy app. When using the following code snippet in a model hook:
```Ember.run.later(this, function() {
      let fooEl = Ember.$('foo.bar');
    }, 1000);```

without this fix crashes the app.

Let me know if the error is not descriptive enough.

cc: @tomdale

### DIFF
--- a/app/initializers/fastboot/error-handler.js
+++ b/app/initializers/fastboot/error-handler.js
@@ -1,0 +1,20 @@
+import Ember from 'ember';
+
+/**
+ * Initializer to attach an `onError` hook to your app running in fastboot. It catches any run loop
+ * exceptions and other errors and prevents the node process from crashing.
+ *
+ */
+export default {
+  name: 'error-handler',
+
+  initialize: function(application) {
+    if (!Ember.onerror) {
+      // if no onerror handler is defined, define one for fastboot environments
+      Ember.onerror = function(err) {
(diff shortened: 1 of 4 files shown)


tomdale:
@kratiahuja Looks great! Do you think you could:

1. Ensure that Ember.onerror doesn't already exist so we don't inadvertently clobber it?
2. Add a test?

Let me know if you need help creating a test app for the fixtures directory.

kratiahuja:
@tomdale Thanks for the feedback
1. Oops, I forgot to take that into picture. Yes, I will add that in. 
2. Aah I wasn't aware of creating a test app in fixture directory existed. I'll add that in tonight/tomorrow.

kratiahuja:
@tomdale Added a simple integration test and the check to not clobber an already defined `Ember.onerror` hook. 

Please let me know if I am missing anything.

kratiahuja:
cc: @danmcclain

danmcclain:
Thanks!

PR #179, opened 2016-04-21
Remove old Nodes from Travis testing

tomdale:


rwjblue:
I would suggest the following:

```
node_js:
  - "stable"
  - "4"
  - "0.12"
```

PR #180, opened 2016-04-21
Implement a function to defer rendering via promise chaining

tomdale:
This is #169 with the merge conflict resolved

### DIFF
--- a/tests/fixtures/async-content/app/components/async-content.js
+++ b/tests/fixtures/async-content/app/components/async-content.js
@@ -1,0 +1,20 @@
+import Ember from 'ember';
+
+export default Ember.Component.extend({
+  fastboot: Ember.inject.service(),
+
+  init() {
+    this._super(...arguments);
+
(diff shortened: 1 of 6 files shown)


tests/fixtures/async-content/app/components/async-content.js
```diff
@@ -0,0 +1,20 @@
...
+export default Ember.Component.extend({
+  fastboot: Ember.inject.service(),
+
+  init() {
+    this._super(...arguments);
+
+    let deferred = Ember.RSVP.defer();
+
+    Ember.run.later(() => {
+      if (!this.get('isDestroyed')) {
```
stefanpenner: this callback should not bet let run if the component has been destroyed. Instead on destroy, we should actually cancel this timer, and more importantly we should also cleanup/resolve the promise.

danmcclain:
#169 has been updated, this is no longer needed

PR #182, opened 2016-04-26
Fix deprecationAlias calls

danmcclain:
Accidentally made an infinite loop in deprecation of
`fastboot.{cookies,headers}`


rwjblue:
Whoops! LGTM

PR #183, opened 2016-04-26
support inclusion as addon dependency

pwfisher:
Similar to https://github.com/tomdale/ember-network/pull/6, build fails if you include an addon which, in turn, includes this addon.


pwfisher:
Other hooks than `included` are needed, so it looks like this must be a fist-level dependency?

pwfisher:
also note `app.app` will not work for 3rd level dependencies https://github.com/ember-cli/ember-cli/issues/5747#issuecomment-206662695

PR #187, opened 2016-05-05
Support storeConfigInMeta, add config to package.json

### LINKED ISSUE #160: why not leave `configInMeta: true`
Seems like the fastboot server could just provide meta before, in-fact meta being configurable might be a great way for fastboot to quickly change options on the fly, without needing a recompile.

danmcclain:
Fixes #160

* Stop forcing `storeConfigInMeta` to false in the browser build
* Store the application config in `package.json`
* Bump `ember-fastboot-server` version to 0.7.3 ([diff from 0.7.2](https://github.com/ember-fastboot/ember-fastboot-server/compare/v0.7.2...v0.7.3))

### DIFF
--- a/index.js
+++ b/index.js
@@ -35,9 +35,12 @@ module.exports = {
    */
   included: function(app) {
     patchEmberApp(app);
+  },
 
-    // We serve the index.html from fastboot-dist, so this has to apply to both builds
-    app.options.storeConfigInMeta = false;
+  config: function() {
+    if (this.app && this.app.options.__is_building_fastboot__) {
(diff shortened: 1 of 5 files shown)


rwjblue:
Looks good. This allows a consuming application's normal `dist/` build to function exactly the same as without fastboot.  

---

We chatted a bit in slack about an even better solution, that would avoid the asset being different based on config (this aids in the ability to promote from one environment to another without rebuilding the assets for example).

rwjblue:
👍 - LGTM

PR #189, opened 2016-05-07
better thennable check

habdelra:
the previous assert for thennable would actually execute the `then()` as part of the assert, which was causing probs


danmcclain:
Huh, I wasn't aware assert actually called the function. TIL

rwjblue:
It does for ember versions prior to 2.6 (with a deprecation).

danmcclain:
@habdelra Danke!

PR #195, opened 2016-05-14
FastBootLocation API

arjansingh:
@rwjblue So this is a first draft. There are a couple of issues:

1. `replaceWith` and `intermediateTransitionTo` die hard. Specifically, `setURL` is not getting the path to redirect to when those two functions are called. Check the failing tests in TravisCI to see those examples.

2. `transitionTo` still renders the body. The middleware or fastboot could be modified to not include it, but a better solution would be to tell Ember to abort the page rendering. Tips on how to do that would be appreciated. I couldn't find anything straightforward in the docs.

If we can solve those, I can add a few more test cases (covering `model`, and `afterModel` hooks come to mind) and get this ready for submission.

### DIFF
--- a/app/locations/none.js
+++ b/app/locations/none.js
@@ -1,0 +1,53 @@
+import Ember from 'ember';
+
+const {
+  computed,
+  computed: { reads },
+  inject: { service },
+  get,
+  getOwner
+} = Ember;
+
+export default Ember.NoneLocation.extend({
+  implementation: 'fastboot',
+  fastboot: service(),
+
+  _fastbootHeadersEnabled: computed(function () {
+    const config = getOwner(this).resolveRegistration('config:environment');
+    return !!get(config, 'fastboot.fastbootHeaders');
+  }),
+
+  _redirectCode: computed(function () {
+    const TEMPORARY_REDIRECT_CODE = 307;
+    const config = getOwner(this).resolveRegistration('config:environment');
+    return get(config, 'fastboot.redirectCode') || TEMPORARY_REDIRECT_CODE;
+  }),
+
+  _response: reads('fastboot.response'),
+  _request: reads('fastboot.request'),
+
+  setURL(path) {
+    if (get(this, 'fastboot.isFastBoot')) {
+      const currentPath = get(this, 'path');
+      const isInitialPath = !currentPath || currentPath.length === 0;
+      const isTransitioning = currentPath !== path;
+      let response = get(this, '_response');
+
+      if (isTransitioning && !isInitialPath) {
+        let protocol = get(this, '_request.protocol');
+        let host = get(this, '_request.host');
+        let redirectURL = `${protocol}://${host}${path}`;
+
+        response.statusCode = this.get('_redirectCode');
+        response.headers.set('location', redirectURL);
+      }
+
+      // for testing and debugging
+      if(get(this, '_fastbootHeadersEnabled')) {
+        response.headers.set('x-fastboot-path', path);
+      }
+    }
+
+    this._super(...arguments);
+  }
+});
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "broccoli-merge-trees": "^1.1.1",
     "broccoli-plugin": "^1.2.1",
     "broccoli-stew": "^1.2.0",
-    "fastboot-express-middleware": "1.0.0-beta.5",
+    "fastboot-express-middleware": "1.0.0-beta.6",
     "ember-cli-babel": "^5.1.5",
     "ember-test-helpers": "^0.5.22",
     "express": "^4.8.5",
--- a/test/fastboot-location-config-test.js
+++ b/test/fastboot-location-config-test.js
@@ -1,0 +1,52 @@
+var chai = require('chai');
+var expect = chai.expect;
+var RSVP = require('rsvp');
+var AddonTestApp = require('ember-cli-addon-tests').AddonTestApp;
+var request = require('request');
+var get = RSVP.denodeify(request);
+
+describe('FastBootLocation Configuration', function () {
+  this.timeout(300000);
+
+  before(function () {
+    app = new AddonTestApp();
+
+    return app.create('fastboot-location-config')
+    .then(function () {
+      return app.startServer({
+        command: 'fastboot',
+        additionalArguments: ['--serve-assets']
+      });
+    });
+  });
+
+  after(function () {
+    return app.stopServer();
+  });
+
+  it('should use the redirect code provided by the EmberApp', function () {
+    return get({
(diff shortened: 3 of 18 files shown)


arjansingh:
Item 1. is caused by:
https://github.com/emberjs/ember.js/blob/45248e2fb2c71ec60bb879c7588beeaeffe6dc6c/packages/ember-application/lib/system/application-instance.js#L300

arjansingh:
Update:

1. `replaceWith` works! Coder error. My test application had a type in the route I was testing. 😛  

2. `intermediateTransitionTo` still has issues. I'm looking for suggestions on how to handle it since it does not call `setURL` or any other Location API functions.

3. `body` still rendering inside of `transitionTo` and `replaceWith`. Any ideas on how to abort rendering? Or should I just delete the result body before returning it to the middleware?

app/locations/none.js
```diff
@@ -0,0 +1,37 @@
+import Ember from 'ember';
+
+const {
+  computed: { reads },
+  inject: { service },
+  get
+} = Ember;
+
+const TEMPORARY_REDIRECT_CODE = 307;
```
marcoow: I think we should just use 302 here. 307 was added so that servers can require subsequent requests after redirect responses to use the same request method as the original request that resulted in the redirect. While we want the same request method to be used of course, 307 was really only added for everythong *other* than `GET` and FastBoot is only ever dealing with `GET` anyway (in fact all other request methods expect for `HEAD` maybe) should actually be ignored by the FastBoot middleware - not sure whether that's actually the case yet.

app/locations/none.js
```diff
@@ -0,0 +1,37 @@
...
+
+      response.statusCode = TEMPORARY_REDIRECT_CODE;
+      response.headers.set('location', redirectURL);
+    }
+
+    response.headers.set('x-fastboot-path', path); // for testing and debugging
+
+    this._super(...arguments);
+  }
+});
```
marcoow: I assume this file is called `none.js` so it overrides the `none` location which automatically gets used in FastBoot? While this is of course pretty convenient it seems a bit implicit - actually monkey-patching our `fastboot` location into the `none` location? Wouldn't it be much clearer to add `fastboot` as an additional location and explicitly change the configuration so that that gets used when running in the server?

app/locations/none.js
danmcclain: 302 is `FOUND`, whereas `307` is `TEMPORARY REDIRECT`, which I'd 307 is more semantically correct. The following note is from [the 302 spec][302spec] (emphasis mine):

> Note: RFC 1945 and RFC 2068 specify that the client is not allowed to change the method on the redirected request.  However, most existing user agent implementations treat 302 as if it were a 303 response, performing a GET on the Location field-value regardless of the original request method. **The status codes 303 and 307 have been added for servers that wish to make unambiguously clear which kind of reaction is expected of the client.**

302 has ambiguity in its handling, whereas 307 is clear
[302spec]: http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.3.3

danmcclain:
As `intermediateTransitionTo` is not a final state (most often used to transition to a loading state), we should probably leave it as is. The page should do another transition before the final rendering, I would think. @rwjblue @tomdale concur?

app/locations/none.js
arjansingh: Agreed on `307 - TEMPORARY REDIRECT`. If we were trying to maintain HTTP 1.0 compatibility `302 - FOUND` would make sense. As it is, `307` is a very clear spec. We could make the code configurable with a `307` default, so people can override it. Would anyone object to that?

app/locations/none.js
arjansingh: I'm not a fan of this either, but @rwjblue and I talked ot over in https://github.com/ember-fastboot/fastboot/issues/54 and that was the cleanest way we found to do it. 

The only other way to accomplish this is to extend `Ember.Router.setupRouter`, which we both thought was messier.

arjansingh:
@danmcclain This issue I'm encountering with `intermediateTransitionTo` right now, is that it is not even transitioning correctly to the final route. 

For its functionality to be maintained, it needs to:

1. Not change the requested `path`
2. Insert the content from the transition into the `body

Currently, 1 happens, but 2 does not. Any thoughts on handeling this? My thought is that if you get an `intermediateTransitionTo` we just return `index.html` and just "slowboot" everything on the client side like a normal Ember app.

app/locations/none.js
marcoow: Making the exact code configurable sounds fine. Didn't want to start an argument about HTTP semantics here - I think that could easily get out of control ;)

arjansingh: WIll do.

app/locations/none.js
marcoow: Couldn't we just change the config when running in FastBoot so that it says

```js
locationType: 'fastboot',
```

?

rwjblue: Indeed we could, but Ember internally **hard codes** the location type when using the `visit` API to `none`.  That is a bug, and we should definitely fix it but we will have to work around that for now.

arjansingh: the [hardcoded lines](https://github.com/emberjs/ember.js/blob/v2.5.0/packages/ember-application/lib/system/application-instance.js#L401-L405)

rwjblue:
@arjansingh and I dug into this a bit today, and have determined that the issue mentioned above is not a blocker for this PR.  Basically, in Ember if you do an intermediate transition during another transition you *must* abort the initial transition.  When discovering that we also uncovered a larger issue WRT aborted transitions rejection and that error being unhandled.

@tomdale - This is good to go from my perspective...
@arjansingh - Can you open an issue for the unhandled TransitionAborted errors?

arjansingh:
Alright updated tests to disable the `intermediateTransitionTo` test. We can modify and reenable it when we resolve #202.

I also will update assertions for the other location tests to check for appropriate body content when https://github.com/ember-fastboot/fastboot/pull/64 is released with the next fastboot beta.

app/locations/none.js
```diff
@@ -0,0 +1,45 @@
...
+      if (isTransitioning && !isInitialPath) {
+        let protocol = get(this, '_request.protocol');
+        let host = get(this, '_request.host');
+        let redirectURL = `${protocol}://${host}${path}`;
+
+        response.statusCode = this.get('_redirectCode');
+        response.headers.set('location', redirectURL);
+      }
+
+      response.headers.set('x-fastboot-path', path); // for testing and debugging
```
danmcclain: I would only display this header when you enable it via a config flag, otherwise we leak what is running

arjansingh: 👍  We can config this.

PR #200, opened 2016-05-25
ember fastboot --watch

pwfisher:
@rwjblue Tests coming soon, just want to open the PR for some early feedback. Does this approach look reasonable?

### DIFF
--- a/lib/commands/fastboot.js
+++ b/lib/commands/fastboot.js
@@ -1,13 +1,19 @@
-var RSVP = require('rsvp');
-var exec = RSVP.denodeify(require('child_process').exec);
+'use strict';
+const RSVP = require('rsvp');
+const getPort = RSVP.denodeify(require('portfinder').getPort);
+const ServerTask = require('../tasks/fastboot-server');
+const SilentError = require('silent-error');
+
+const blockForever = () => (new RSVP.Promise(() => {}));
+const noop = function() { };
 
 module.exports = {
   name: 'fastboot',
-  description: 'Runs a server to render your app using FastBoot.',
+  description: 'Builds and serves your FastBoot app, rebuilding on file changes.',
 
   availableOptions: [
     { name: 'build', type: Boolean, default: true },
-    { name: 'watch', type: Boolean, default: false, aliases: ['w'] },
+    { name: 'watch', type: Boolean, default: true, aliases: ['w'] },
     { name: 'environment', type: String, default: 'development', aliases: ['e',{'dev' : 'development'}, {'prod' : 'production'}] },
     { name: 'serve-assets', type: Boolean, default: false },
     { name: 'host', type: String, default: '::' },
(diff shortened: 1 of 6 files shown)


danmcclain:
I'd argue that `ember fastboot` should watch by default like `ember serve` does

pwfisher:
I'm thinking `SIGHUP` would be more appropriate than `SIGUSR1`.

> For this reason, applications that did not require a controlling terminal, such as daemons, would re-purpose SIGHUP as a signal to re-read configuration files, or reinitialize. This convention survives to this day in packages such as Apache and Sendmail. [[Wikipedia](https://en.wikipedia.org/wiki/SIGHUP)]

lib/commands/fastboot.js
```diff
@@ -15,27 +16,29 @@ module.exports = {
...
-  runCommand: function(appName, options) {
-    var commandOptions = this.commandOptions;
-    var outputPath = commandOptions.outputPath;
-    var assetsPath = commandOptions.assetsPath;
+  startServer: function(options) {
     var ui = this.ui;
+    var self = this;
 
-    ui.writeLine("Installing FastBoot npm dependencies");
+    if (this._fastBootListener) {
```
jasonmit: I recall some issues in the past where programmatically restarting a server required I also flush require.cache otherwise stale modules were loaded.  I could be wrong, just jotting down something that came to mind

jasonmit: Ah, now I remember where this was an issue: https://github.com/ember-cli/ember-cli/blob/da4749d3b27a2ce7c529de0d07c0ac3e702e7d47/lib/tasks/server/express-server.js#L238-L250

pwfisher: Thanks, I'm adding that.

pwfisher: The issue would affect `dist/fastboot/node_modules`, which we `require` in  [`buildWhitelistedRequire`](https://github.com/ember-fastboot/fastboot/blob/f1b311d/lib/ember-app.js#L92)

jasonmit: 👍

pwfisher:
FastBoot Command

run
- check for early exit
- run build task, if any
- run server task
- block forever

FastBoot Command tests

when port is 0
- find an open port
- call tasks.FastbootServer.run with found port

when port > 0 is in use
- exit early with SilentError

when build and watch enabled
- call tasks.BuildWatch.run

when build and not watch enabled
- call tasks.Build.run
- not call tasks.BuildWatch.run

when build not enabled
- not call tasks.Build.run
- not call tasks.BuildWatch.run

FastBoot Server Task

run
- listen for SIGHUP
- start server

FastBoot Server Task tests

when run is called
- run server on port
- call reload on SIGHUP

when starting server
- run npm install in server root
- require server deps

when reload is called
- destroy sockets
- close server
- clear server require cache
- start server

when reload is called multiple times
- not try to destroy old sockets
- pick up changes in npm requires

when SIGHUP received
- call outputReady

when outputReady while server is starting
(e.g. app file change during server npm install)
- wait on start, then reload

when outputReady multiple times during startup
(e.g. fast app build, slow server npm install)
- call reload only once

leo:
@danmcclain Great idea! And it should be disabled if the `--environment` flag is set to "production".

stefanpenner:
I would really like it if fastboot and normal serve could be united during development. An early discussion I started about making app.js the same regardless of fastboot or browser would nicely enable this.

danmcclain:
@stefanpenner We actually talked about this during the last FastBoot Sync meeting

lib/tasks/fastboot-server.js
```diff
@@ -0,0 +1,121 @@
...
+  },
+
+  restart(options) {
+    if (this.restartPromise) {
+      debug('schedule immediate restart');
+      this.restartAgain = true;
+      return;
+    }
+    debug('restart');
+    this.restartPromise = this.stop()
```
rwjblue: I think we should try to avoid stopping and starting the server for each reload. With some tweaks to our usage of fastboot-express-middleware we should be able to make this `restartPromise` closer to:

```js
this.restartPromise = RSVP.resolve()
  .then(() => this.clearRequireCache(options.outputPath))
  .then(() => this._fastboot.reload())
  .finally(() => { /* snip from existing */ })
```

lib/tasks/fastboot-server.js
```diff
@@ -0,0 +1,121 @@
...
+    process.on('SIGHUP', restart);
+  },
+
+  start(options) {
+    debug('start');
+    this.ui.writeLine('Installing FastBoot npm dependencies');
+
+    return this.exec('npm install', { cwd: options.outputPath })
+      .then(() => {
+        const middleware = this.require('fastboot-express-middleware')(options.outputPath);
```
rwjblue: We should change how we invoke `fastboot-express-middleware` to make reloading easier.  We should basically copy from how it is done in `fastboot-app-server`:

* https://github.com/ember-fastboot/fastboot-app-server/blob/master/lib/worker.js#L52-L61 - Setting up the middleware.
* https://github.com/ember-fastboot/fastboot-app-server/blob/master/lib/worker.js#L41 - To reload/restart.

rwjblue:
I left a few inline comments, but overall this looks very good.  

Also, stylistic nit RE: `const`: I'd prefer to stay in line with Ember and only use `const` in module scope.

pwfisher:
Per discussion, we will leave the `npm install` on restart since it will "just work" and potential optimization will be complex (`dist/node_modules` is nuked by the build and the naïve solution would rely on `require` finding its dependencies up the directory tree—but this is not guaranteed to work). An optimized restart would be a separate PR.

`const` usage can land as-is and a separate PR will add and apply the linter rules from Ember.

lib/tasks/fastboot-server.js
pwfisher: Faster restart will be a separate task.

lib/tasks/fastboot-server.js
pwfisher: Faster restart will be a separate task.

danmcclain:
Homu has been dead for days 😢

meteormanaged:
Has this been released or just merged in?

PR #201, opened 2016-05-26
Do not compute FastBootService.request in the browser

arjansingh:
@danmcclain Whenever `FastBootService.request` was accessed in the browser, it was not handling `_fastbootInfo.request` being `undefined`. This fix ensures that we return `null` in that case just like `Ember.computed.readOnly` would.

### DIFF
--- a/app/services/fastboot.js
+++ b/app/services/fastboot.js
@@ -84,6 +84,7 @@ export default Ember.Service.extend({
   response: readOnly('_fastbootInfo.response'),
 
   request: computed(function() {
+    if (!get(this, 'isFastBoot')) return null;
     return RequestObject.create({ request: get(this, '_fastbootInfo.request') });
   }),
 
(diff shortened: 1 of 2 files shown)


app/services/fastboot.js
```diff
@@ -1,7 +1,7 @@
 /* global FastBoot */
 import Ember from "ember";
 
-const { deprecate, computed, get } = Ember;
+const { deprecate, computed, get, set } = Ember;
```
danmcclain: Where is `set` being used?

arjansingh: nowhere after my last refactor. good catch!

arjansingh:
Updated to remove `set`

arjansingh:
@danmcclain any other blockers to merging?

tests/unit/services/fastboot-test.js
```diff
@@ -0,0 +1,21 @@
+import { moduleFor, test } from 'ember-qunit';
+
+moduleFor('service:fastboot', 'Unit | Service | fastboot in the browser', {
+  // Specify the other units that are required for this test.
```
danmcclain: Don't need these comment lines

arjansingh: 😁

PR #203, opened 2016-05-29
Specify localhost as default host

leo:
IMHO, "localhost" or "127.0.0.1" should be the default address that gets used when serving an ember app using FastBoot (instead of "::", which allows connections on any IPv6/IPv4 address by default).

Devs will still be able to specify a different host (or "::" for all) using `--host`.

**BEFORE**
```
Ember FastBoot running at http://[::]:3000
```

**AFTER**

```
Ember FastBoot running at http://127.0.0.1:3000
```

By the way, I've also considered simply adding an if-statement to the method that logs the message which replaces the host with "localhost" if the original value is "::". But in my opinion, that's not a good solution because "::" allows connections over all addresses, not just the localhost one.

A nice side-effect of this change is that the user will be able to click the address within the command line to open it in the browser using `doubleclick` + `cmd`.

Besides of that, it might also be a little more secure because the app will only be available on the exact address shown in the logged message.

### DIFF
--- a/lib/commands/fastboot.js
+++ b/lib/commands/fastboot.js
@@ -9,7 +9,7 @@ module.exports = {
     { name: 'build', type: Boolean, default: true },
     { name: 'environment', type: String, default: 'development', aliases: ['e',{'dev' : 'development'}, {'prod' : 'production'}] },
     { name: 'serve-assets', type: Boolean, default: false },
-    { name: 'host', type: String, default: 'localhost' },
+    { name: 'host', type: String, default: '::1' },
     { name: 'port', type: Number, default: 3000 },
     { name: 'output-path', type: String, default: 'dist' },
     { name: 'assets-path', type: String, default: 'dist' }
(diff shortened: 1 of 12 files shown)


lib/commands/fastboot.js
```diff
@@ -9,7 +9,7 @@ module.exports = {
     { name: 'build', type: Boolean, default: true },
     { name: 'environment', type: String, default: 'development', aliases: ['e',{'dev' : 'development'}, {'prod' : 'production'}] },
     { name: 'serve-assets', type: Boolean, default: false },
-    { name: 'host', type: String, default: '::' },
+    { name: 'host', type: String, default: 'localhost' },
```
danmcclain: Let's have this be `::1` so that it is still IPv6

leo: Done!

leo:
@danmcclain Looks like the commit broke the tests...?

danmcclain:
Looks like it will no longer support IPv4 127.0.0.1?

leo:
@danmcclain Replaced all `localhost`s and `127.0.0.1`s with `::1`, still doesn't seem to work.

PR #217, opened 2016-06-15
Fix deprecation warning for using EmberApp.concatFiles

### LINKED ISSUE #214: Avoid using private API - EmberApp.concatFiles (RFC needed)
Removing this addon clears many deprecation warnings in Ember CLI 2.7 beta for using EmberApp.concatFiles.

I think there is some private option that allows you to call concatFiles without a deprecation warning.

Gaurav0:
Fixes #214

### DIFF
--- a/lib/ext/patch-ember-app.js
+++ b/lib/ext/patch-ember-app.js
@@ -21,7 +21,7 @@ function patchEmberApp(emberApp) {
     if (options.annotation === 'Concat: App') {
       tree = this.addonPreconcatTree(tree);
     }
-    return originalConcatFiles.call(this, tree, options);
+    return originalConcatFiles.apply(this, arguments);
   };
 }
 


rwjblue:
This fix seems good for now, however we **must** make an effort to submit an RFC for the overrides we need here.  It is highly likely that at some point in the future, after the `concatFiles` deprecation period is over, that this will just be removed.  If we do not submit an ember-cli RFC and move our needs upstream, we will be up 💩  creek...

lib/ext/patch-ember-app.js
```diff
@@ -17,11 +17,11 @@ function patchEmberApp(emberApp) {
   // and, if it detects that it's a concat for the app tree, invokes our
   // preconcat hook. Afterwards, we invoke the original implementation to
   // return a tree concating the files.
-  emberApp.concatFiles = function(tree, options) {
+  emberApp.concatFiles = function(tree, options, secretArg) {
     if (options.annotation === 'Concat: App') {
       tree = this.addonPreconcatTree(tree);
     }
-    return originalConcatFiles.call(this, tree, options);
+    return originalConcatFiles.call(this, tree, options, secretArg);
```
rwjblue: Can you change this to:

```js
return originalConcatFiles.apply(this, arguments);
```

Gaurav0:
@rwjblue comment addressed

PR #220, opened 2016-06-21
Use fastbootAppConfig.htmlFile if present

bantic:
This change allows someone to add a `fastboot: { htmlFile: 'some-other-file.html' }` section to their ember app's `config/environment.js` and ember-cli-fastboot will use it instead of the default `index.html` file. The use case I'm aiming for is using fastboot to rember an ember app's body into a different HTML shell — for instance Google's AMP requires its own special HTML boilerplate, so in order to use ember-cli-fastboot to render a valid AMP page, it must insert the app's result into a different HTML file.

Is this a viable method to achieve that goal? If so I can add some tests to this PR.

### DIFF
--- a/lib/broccoli/fastboot-config.js
+++ b/lib/broccoli/fastboot-config.js
@@ -17,6 +17,12 @@ function FastBootConfig(inputNode, options) {
   this.outputPaths = options.outputPaths;
   this.appConfig = options.appConfig;
   this._fileToChecksumMap = {};
+
+  if (this.fastbootAppConfig && this.fastbootAppConfig.htmlFile) {
+    this.htmlFile = this.fastbootAppConfig.htmlFile;
+  } else {
+    this.htmlFile = 'index.html';
+  }
 }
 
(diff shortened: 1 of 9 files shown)


tomdale:
@bantic Hey, this looks good. Would you mind adding a test and rebasing?

bantic:
@tomdale Thanks! I've rebased and added tests:
  * that the manfiest is built with the value of `ENV.fastboot.htmlFile` 
  * that the rendered content is interpolated into the custom html file

Should I update the README with information about the `htmlFile` config option? I can add a section before or after the [Disabling incompatible dependencies](https://github.com/ember-fastboot/ember-cli-fastboot#disabling-incompatible-dependencies) section, perhaps?

tomdale:
@bantic: Awesome, thank you! We're moving most of the README content to the website (see https://github.com/ember-fastboot/fastboot-website/pull/41), would you mind adding docs there?

bantic:
Yep, I will add docs there. Thanks!

PR #222, opened 2016-06-24
server task's start shouldn't resolve until listening

ef4:
I was glad to see the new fastboot-server Task, and I started experimenting with using it in a test harness. In the process I noticed this asynchrony bug.

Without this you can't reliably wait until the server is listening before sending in a request or telling it to stop.


rwjblue:
:+1:

Issue #226, opened 2016-07-18
ember-fastboot breaks live css reloading


meteormanaged:
Working with ember-fastboot and ember-cli.  It appears (with some help in the ember slack community) that the ember-fastboot-cli add-on breaks live css reloading.  

Upon installing the add-on, all css changes forces a full page page reload instead of the live css reload, when developing under `ember s`.  Removing the add-on and running it shows the live css reloading working again.

ember-cli: 2.6.3
ember-cli-fastboot: 1.0.0-beta.4

davidpett:
looked into this a bit more. It looks like https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L93 is the start of the problem, diving deeper in the stack https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/lib/broccoli/fastboot-config.js#L34 is the root of the issue. removing that line allows `ember s` to hot reload the CSS, but will break when running any `ember fastboot` commands.

Should any of these hooks be run when not in fastboot mode? I tried to block them with a check of `process.env.EMBER_CLI_FASTBOOT` but it seems to be called a few times, the last one returning `undefined`, so it seems like this check is not reliable at this point.

kpfefferle:
Are this issue and #212 the same issue?

meteormanaged:
Indeed, it appears to be.

stefanpenner:
looks like a dupe, lets close

tomdale:
So ironically, when working on this, @rwjblue and I thought "writing this config file is so fast, and caching has its own overhead, let's just do the naive thing and write it every time."

However, it looks like Ember CLI diffs the output between builds and, if the only change is CSS, does the hot reloading. This Broccoli plugin writes every time, meaning a JSON file gets added to every diff, invalidating the "hot reload."

The right thing to do is probably hash the config and only write if it's different from the last run?

stefanpenner:
@tomdale yup. def something we should make more "fall into a pit of success"

stefanpenner:
@tomdale an example: https://github.com/ef4/ember-browserify/blob/master/lib/stub-generator.js#L64

davidpett:
I'll take a look at that example and update my PR accordingly. This makes much more sense

davidpett:
so there is a dilemma, i got this working in `1.0.0-beta.7`, but `master` doesn't seem to catch any file changes when running `ember s`. any thoughts on what i should do?

PR #230, opened 2016-07-21
block postprocessTree when running ember serve

davidpett:
fixes https://github.com/ember-fastboot/ember-cli-fastboot/issues/226

### DIFF
--- a/index.js
+++ b/index.js
@@ -86,7 +86,8 @@ module.exports = {
    * adds the `fastboot-config.json` file to the root.
    */
   postprocessTree: function(type, tree) {
-    if (type === 'all') {
+    // Only merge `fastbootTree` when not running `ember serve`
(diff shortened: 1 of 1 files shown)


stefanpenner:
This most likely is fixing a symptom not the problem. Why does this cause output to be invalidated?

I believe the root is still that we have to create versions of the app, rather then just 1 version that can be used by web/node/etc.etc.

davidpett:
I agree that it is not fixing the problem, but unsure how to do it. I trace it back to `fs.writeFileSync` here https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/lib/broccoli/fastboot-config.js#L34 that will trigger the page to reload instead of just the styles.

With the current state of single build, is there another solution?

davidpett:
working on a new one

PR #237, opened 2016-07-27
Better `ember fastboot` error message

tomdale:
A previous commit caused stack traces to not be printed if there was an error while loading an app, making debugging much harder.

This commit reintroduces the stack trace to the logged output. It also tries to parse the stack trace and provide more helpful output for people who are not familiar with Ember's internals.

Example with exception from an addon:
1 ember fastboot node 2016-07-27 15-52-44

Example with exception from the app:
fastboot-test zsh 2016-07-27 15-48-37

### DIFF
--- a/lib/tasks/fastboot-server.js
+++ b/lib/tasks/fastboot-server.js
@@ -97,11 +97,11 @@ module.exports = CoreObject.extend({
     this.restartPromise = this.stop()
       .then(() => this.clearRequireCache(options.outputPath))
       .then(() => this.start(options))
-      .catch(e => this.ui.writeLine(e))
+      .catch(e => this.printError(e))
       .finally(() => {
         this.restartPromise = null;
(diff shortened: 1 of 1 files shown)


rwjblue:
This looks great, and a definite improvement! 

I think a future revision could even use sourcemap information to help narrow things down even further.

lib/tasks/fastboot-server.js
```diff
@@ -120,5 +120,58 @@ module.exports = CoreObject.extend({
         delete require.cache[key];
       }
     });
+  },
+
+  /*
+   * Try to show a useful error message if we're not able to start the user's
+   * app in FastBoot.
+   */
+  printError: function(e) {
```
stefanpenner: this function looks pretty complicated, it is likely a good idea to have unit tests covering it.

PR #238, opened 2016-07-27
Use outputReady instead of postBuild

jasonmit:
Based on Robert's advice on slack.

/cc @rwjblue @tomdale


rwjblue:
👍

davidpett:
@rwjblue and @jasonmit this breaks livereloading entirely. any change in the app JS/HBS/CSS does not trigger a rebuild

Issue #212, opened 2016-06-13
CSS Updates cause page reload (navigation) instead of hot reload


rjfranco:
When ember-cli-fastboot is included in a project, css updates cause a page refresh, not sure where in particular this originates, can be reproduced in a new 2.6.1 project where the only step taken is running `ember install ember-cli-fastboot`

tomdale:
I'm not familiar with how CSS hot reloading works in Ember CLI or how we might be interfering with it. That said, I think once we integrate better into the Ember CLI build pipeline by using public hooks (and avoiding a double build of the app) this should hopefully get resolved.

stefanpenner:
@tomdale css hot-reloading, diffs the output of the build, and reports all changed paths via websocket to livereload.js. If hot-reloading is broken, it is most likely because fast-boot causes everything to invalidate/rebuild and change.

stefanpenner:
For reference, the relevant code in ember-cli  is: here https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/server/livereload-server.js#L140-L153

stefanpenner:
There is a debug, [here](https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/server/livereload-server.js#L155) So running the following, will help inform further what is up:

```
`DEBUG="ember-cli:live-reload:*" ember <w/e command>`

kpfefferle:
@rjfranco I think this issue can be closed now that #239 has landed. Confirm?

PR #239, opened 2016-07-27
fix fastboot styles reloading

davidpett:
fixes https://github.com/ember-fastboot/ember-cli-fastboot/issues/226 and https://github.com/ember-fastboot/ember-cli-fastboot/issues/212


tomdale:
This looks good to me, but @stefanpenner @rwjblue what do you think?

PR #244, opened 2016-08-01
Bump Ember CLI to 2.7.0

tomdale:


stefanpenner:
may want to use: https://greenkeeper.io/ I find its a big time saver.

PR #249, opened 2016-08-04
Remove unused start- and destroy-app helpers from mocha `test/` dir

bantic:
These do not appear to actually be used.


stefanpenner:
they can always be brought back

PR #250, opened 2016-08-04
Better shoebox API

ef4:
This adds a new higher-level API for the shoebox.

```js
  model() {
    let fastboot = this.get('fastboot');
    return RSVP.hash({
      // In this example, we generate a random number once on the server,
      // and then the  client app will receive the same number.
      random: fastboot.withShoebox('my-random-number', () => Math.random())
    });
  }
```

`withShoebox` acts as a memoization service. On the server, the given callback will run and generate a value, which will be saved in the shoebox and then returned. On the client, the callback will not run if the value is already present in the shoebox.

`withShoebox` returns a promise, because the callback is allowed to return a promise.

### DIFF
--- a/app/services/fastboot.js
+++ b/app/services/fastboot.js
@@ -1,8 +1,9 @@
 /* global FastBoot */
 import Ember from "ember";
 
-const { deprecate, computed, get } = Ember;
+const { deprecate, computed, get, RSVP } = Ember;
 const { deprecatingAlias, readOnly } = computed;
+const { Promise } = RSVP;
 
 const RequestObject = Ember.Object.extend({
   init() {
@@ -57,6 +58,15 @@ const Shoebox = Ember.Object.extend({
     this.set(key, shoeboxItem);
 
     return shoeboxItem;
+  },
+
+  has(key) {
+    if (this.get('fastboot.isFastBoot')) {
+      let shoebox = this.get('fastboot._fastbootInfo.shoebox');
+      return shoebox && shoebox.hasOwnProperty(key);
+    } else {
+      return !!document.querySelector(`#shoebox-${key}`);
+    }
   }
 });
 
@@ -95,5 +105,21 @@ export default Ember.Service.extend({
   deferRendering(promise) {
     Ember.assert('deferRendering requires a promise or thennable object', typeof promise.then === 'function');
     this._fastbootInfo.deferRendering(promise);
+  },
+
+  withShoebox(key, fn) {
+    return new Promise((resolve, reject) => {
+      let shoebox = this.get('shoebox');
+      if (shoebox.has(key)) {
+        resolve(shoebox.retrieve(key));
+        return;
+      }
+      Promise.resolve(fn()).then(value => {
+        if (this.get('isFastBoot')) {
+          shoebox.put(key, value);
+        }
+        return value;
+      }).then(resolve, reject);
+    });
   }
 });


app/services/fastboot.js
```diff
@@ -95,5 +105,21 @@ export default Ember.Service.extend({
...
+  },
+
+  withShoebox(key, fn) {
+    return new Promise((resolve, reject) => {
+      let shoebox = this.get('shoebox');
+      if (shoebox.has(key)) {
+        resolve(shoebox.retrieve(key));
+        return;
+      }
+      Promise.resolve(fn()).then(value => {
```
stefanpenner: `resolve(Promise.resolve(fn())...` not `then(resolve, reject)`

danmcclain:
I love this API @ef4

app/services/fastboot.js
```diff
@@ -57,6 +58,15 @@ const Shoebox = Ember.Object.extend({
...
 
     return shoeboxItem;
+  },
+
+  has(key) {
+    if (this.get('fastboot.isFastBoot')) {
+      let shoebox = this.get('fastboot._fastbootInfo.shoebox');
+      return shoebox && shoebox.hasOwnProperty(key);
+    } else {
+      return !!document.querySelector(`#shoebox-${key}`);
```
stefanpenner: if shoebox is for random data (that we likely don't want displayed) should we use meta or script tags?

stefanpenner: that could also make them addressible via `import`

danmcclain: Meta tags have limits on length that differ from browser to browser @stefanpenner

stefanpenner: ah so big datasets

arjansingh:
So this immediately solves some problems I'm running into. What else (besides tests) needs to be done to get this in?

steveszc:
It seems like the problem `withShoebox` is trying to solve is preventing the callback from running twice on initial pageload (once on the server, once when the client loads). However the API seems to prevent the callback from ever running on the client if the shoebox holds data matching the requested key. This seems undesirable.

I see a common use case where you only want the client to access the shoebox data one time (on initial payload) and then use the callback every subsequent time the data is requested.  

Maybe it would make sense to add an argument flag or a separate API that would implement this idea of "single use" shoebox data?

BenKingBBC:
@steveszc it's my understanding that `shoebox.retrieve` drops the value from the shoebox after you have retrieved it, which should mean what you're suggesting isn't an issue.

steveszc:
@BenKingBBC That hasn't been my experience using `shoebox.retrieve` in my app, and I don't see anything happening in the [method's source code](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/app/services/fastboot.js#L40) that suggests that it removes values from the shoebox after they're retrieved. 

It does look like it caches the value after it is retrieved from the DOM, but subsequent calls to shoebox.retrieve for a given key would just get the cached value instead of going back to the DOM shoebox.

I'm happy to open a separate issue if there is consensus that "single-use" use case for shoebox values should be handled by `shoebox.retrieve` instead of a higher-level API.

danmcclain:
@ef4 We discussed as a team, and like and approve the API, would you have bandwidth to write tests so that we can land this?

vladucu:
this would be really awesome!

is there anything else blocking this except tests? or latest `rc2` release also affects it?

app/services/fastboot.js
```diff
@@ -95,5 +105,21 @@ export default Ember.Service.extend({
...
+  },
+
+  withShoebox(key, fn) {
+    return new Promise(resolve => {
+      let shoebox = this.get('shoebox');
+      if (shoebox.has(key)) {
+        resolve(shoebox.retrieve(key));
+        return;
+      }
+      resolve(Promise.resolve(fn()).then(value => {
```
bekzod: do you need wrapping `Promise.resolve` ? I think it will work without it too

xg-wang:
cc/ @suchitadoshi1987

suchitadoshi1987:
This is a good one to work on and get it in. I ll pick this one up if @ef4 is not working actively on it.

PR #253, opened 2016-08-09
Enable gzip compression for the express server

zeppelin:
Same as in https://github.com/ember-cli/ember-cli/commit/a11c4af6e599fe1149328463b7e0a6eba9696c2d


danmcclain:
This will only enable compression in development mode (as you would use something like `fastboot-app-server`).

LGTM though

rwjblue:
Ya, seems good to me.

PR #257, opened 2016-09-06
Start the FastBoot server immediately when --build=false

treywood:
Going through the User Guide, I noticed that using the suggested command for debugging:

```
$ ember build && node --debug-brk ./node_modules/.bin/ember fastboot --no-build
```

would never start the server. After poking around I found that `$ ember fastboot --no-build` would start and immediately exit since no 'postBuild' event is fired to start the server. The changes in this PR cause the server to start immediately when --build=false


stefanpenner:
> ./node_modules/.bin/ember fastboot --no-build

the following command seems more aligned with existing (similar features)

```
ember test --path /path/to/dist/to/run/tests/against
ember fastboot --path /path/to/dist/to/serve // <-- recommendation
```

tomdale:
@treywood Hey, thanks for the PR!

We discussed this PR on the call today. I am a little leery about accepting this patch as-is. There is some work on-going to give addons the ability to patch into the built-in Ember Express server. That seems like a better option than hard-coding in special case behavior about the build into the server. (IMO the server _already_ shouldn't have any knowledge of the build system, and I don't want to canonize that bad architectural system by building on it further.)

That said, I think I would be OK accepting a PR in the interim if there was a _different_ event than `postBuild` that we could listen for that was triggered whether or not the build was kicked off—although I'm not sure if such a thing exists. But I'm not comfortable with the server starting to parse build-specific CLI options.

simonihmig:
Just found this by coincidence. Submitted a PR for that issue recently that was already merged: https://github.com/ember-fastboot/ember-cli-fastboot/pull/346. Should be able to close this then!

rwjblue:
Dang, I'm sorry about missing this PR originally @treywood.

PR #260, opened 2016-09-08
[fix] package-json-test.js

arjansingh:
I just tried to run this test locally and it failed since we have:

```
        "foo": "1.2.3",
        "bar": "^7.8.9",
        "baz": "0.2.0",
```

in the fixture app's `package.json`.


arjansingh:
Ok that seems to do harm in CI. Nevermind.

PR #262, opened 2016-09-10
Handle fingerprinted asset maps

arjansingh:
[broccoli-asset-rev](https://github.com/rickharrison/broccoli-asset-rev) has the option to fingerprint your asset map file.

This PR adds support for that scenario.

I'm not happy about using the wild card, but I don't think a `RegExp` will work there.

### DIFF
--- a/lib/broccoli/fastboot-config.js
+++ b/lib/broccoli/fastboot-config.js
@@ -98,13 +98,17 @@ FastBootConfig.prototype.buildDependencies = function() {
 };
 
 FastBootConfig.prototype.readAssetManifest = function() {
+  var ui = this.ui;
   var assetMapPath = path.join(this.inputPaths[0], 'fastbootAssetMap.json');
+  var assetMapEnabled = this.assetMapEnabled;
 
   try {
     var assetMap = JSON.parse(fs.readFileSync(assetMapPath));
     return assetMap;
   } catch (e) {
-    // No asset map was found, proceed as usual
+    if (this.assetMapEnabled) {
+      ui.writeLine(fmt("fastbootAssetMap.json not found at: %s", assetMapPath), ui.WARNING);
+    }
   }
 };
 
--- a/lib/broccoli/fastboot-build.js
+++ b/lib/broccoli/fastboot-build.js
@@ -24,6 +24,8 @@ function FastBootBuild(options) {
   })[0];
 
   if (assetRev && assetRev.options) {
+    this.assetMapEnabled = !!(assetRev.options.enabled && assetRev.options.assetMapPath);
+
     if (assetRev.options.assetMapPath) {
       this.assetMapPath = assetRev.options.assetMapPath;
     }
@@ -118,7 +120,7 @@ FastBootBuild.prototype.buildConfigTree = function(tree) {
   return new FastBootConfig(tree, {
     project: this.project,
     name: this.app.name,
-    assetMapPath: this.assetMapPath,
+    assetMapEnabled: this.assetMapEnabled,
     outputPaths: this.app.options.outputPaths,
     ui: this.ui,
     fastbootAppConfig: fastbootConfig,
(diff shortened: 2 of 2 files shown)


lib/broccoli/fastboot-config.js
```diff
@@ -104,7 +104,9 @@ FastBootConfig.prototype.readAssetManifest = function() {
     var assetMap = JSON.parse(fs.readFileSync(assetMapPath));
     return assetMap;
   } catch (e) {
-    // No asset map was found, proceed as usual
+    if (assetMapPath) {
```
arjansingh: We should at least warn about this when you have an asset map.

danmcclain: The warning should only show if the assets are fingerprinted and there is not asset map, right? That would be a bad state. If they aren't fingerprinted, then they would be at their expected paths, and the asset map isn't required (As least that's my understanding)

arjansingh: I thought about that. Any time getting an `assetMap` load fails we should log it. If it's a malformed `assetMap` that fails `parse` or if it's just absent, the user would benefit from knowing that.  We can print the error too.

danmcclain: I think it may be beneficial to include the error message and also state that it's warning, rather than an error in the case that it's not required (non-fingerprinted assets)

arjansingh: I can do that.

lib/broccoli/fastboot-config.js
```diff
@@ -98,13 +98,16 @@ FastBootConfig.prototype.buildDependencies = function() {
...
+  var ui = this.ui;
   var assetMapPath = path.join(this.inputPaths[0], 'fastbootAssetMap.json');
 
   try {
     var assetMap = JSON.parse(fs.readFileSync(assetMapPath));
     return assetMap;
   } catch (e) {
-    // No asset map was found, proceed as usual
+    if (assetMapPath) {
+      ui.writeLine(fmt("FastBoot Build: No asset map was found at: %s", assetMapPath), ui.WARNING);
```
arjansingh: @danmcclain switched to using `this.ui` as well.

arjansingh:
Good to merge?

briangonzalez:
Any luck reviewing @danmcclain @tomdale?

lib/broccoli/fastboot-config.js
tomdale: IMO, you should only emit warnings if they are actionable. In this case, it will always emit a warning in development mode that you can't get rid of, because asset rewriting is not turned on in development mode. That would drive me crazy, and it will make people less likely to see other, *real* warnings that we might emit in the future.

tomdale: In other words, I'm in favor of emitting a warning or error if we can't parse the file, or if we would expect the file to exist and it doesn't, but I am not in favor of logging a warning always in development mode if fingerprinting is turned off (which it is by default).

arjansingh: This warning (especially after taking care of @danmcclain's feedback) will only trigger if asset map generation is enabled and no asset map was found. I think that is absolutely a scenario that warrants a warning.

tomdale: Per the discussion at this week's meeting:

This warning should only be emitted if `assetRev && assetRev.options && assetRev.options.enabled`.

arjansingh: @tomdale changed made in https://github.com/ember-fastboot/ember-cli-fastboot/pull/262/commits/edcd3af9b284e67cf6c1ae9e46a694a3bf98b5d6

lib/broccoli/fastboot-config.js
```diff
@@ -98,13 +98,16 @@ FastBootConfig.prototype.buildDependencies = function() {
 };
 
 FastBootConfig.prototype.readAssetManifest = function() {
+  if (!this.assetMapEnabled) return '';
```
arjansingh: Since we know this before the try/catch block I decided to leverage it to do an early exit. Let me know if you'd rather I didn't.

arjansingh: Nevermind this comment.

PR #269, opened 2016-09-26
Only start app once, call fastboot.reload

danmcclain:
Remove `npm install` from watch cycle

As @ef4 pointed out, it's not required, as the node_modules in the root
of the project will be used

Need to update tests

### DIFF
--- a/index.js
+++ b/index.js
@@ -139,6 +139,10 @@ module.exports = {
     return fastbootBuild.toTree();
   },
 
+  outputReady: function() {
+    this.emit('outputReady');
+  },
+
   postBuild: function() {
     this.emit('postBuild');
   },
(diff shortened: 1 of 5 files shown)


stefanpenner:
> Remove npm install from watch cycle
> 
> As @ef4 pointed out, it's not required, as the node_modules in the root
> of the project will be used

👍

rwjblue:
Rebased against recent updates to package.json by @kellyselden (which included ember-addon-tests bump)

rwjblue:
:confused:

josemarluedke:
Hey @danmcclain could you outline what changes in the tests needs to happen so we can land this PR? I'm happy to work in updating these tests if I have more information and a path forward.

Thanks.

danmcclain:
@josemarluedke the tests are updated, we just need to conquer Travis, @rwjblue promised me he'd dig in, just need to bug him :troll:

PR #273, opened 2016-10-06
In Ember 2.7+ assets should be served from a rootURL-prefixed path

kfoisy:
In Ember 2.7+, using the baseURL has been deprecated (see http://emberjs.com/blog/2016/04/28/baseURL.html) and the assets should be served from a rootURL prefixed path.


wytlytningNZ:
Bump please 👍

varoot:
What's the status of this PR?

stefanpenner:
I believe this is now defunct as master and soon 1.0 just piggy back on `ember s`

PR #281, opened 2016-10-18
Update the version of Ember that doesn't need the dom helper patches.

kratiahuja:
Since glimmer2 is not going to be shipped with Ember 2.9, we need the dom helper patches for rendering apps on server side.


rwjblue:
```
npm publish
+ ember-cli-fastboot@1.0.0-beta.10
```

PR #284, opened 2016-10-22
Get ember version from ember-source if there is no ember in bower

josemarluedke:
This allow users to use ember from npm.


rwjblue:
Thanks @josemarluedke!

PR #300, opened 2016-12-09
WIP: Register fastboot and ember-cli-fastboot versions

danmcclain:
Allows fastboot server to print out current versions on app boot

### DIFF
--- a/addon/versions.js
+++ b/addon/versions.js
@@ -1,0 +1,6 @@
+import Ember from 'ember';
+
+if (Ember.libraries) {
+  Ember.libraries.register('FastBoot', 'FASTBOOT_VERSION');
+  Ember.libraries.register('Ember CLI FastBoot', 'EMBER_CLI_FASTBOOT_VERSION');
+}
--- a/index.js
+++ b/index.js
@@ -2,9 +2,12 @@
 'use strict';
 
 var path = require('path');
+var fs = require('fs');
 
 var EventEmitter = require('events').EventEmitter;
+var getGitInfo = require('git-repo-info');
 var mergeTrees = require('broccoli-merge-trees');
+var replace = require('broccoli-string-replace');
 var VersionChecker = require('ember-cli-version-checker');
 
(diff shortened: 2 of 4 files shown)


danmcclain:
Output:
```
DEBUG: -------------------------------
DEBUG: Ember              : 2.7.3
DEBUG: Ember Data         : 2.8.1
DEBUG: FastBoot           : 1.0.0-rc.0
DEBUG: Ember CLI FastBoot : 1.0.0-beta.13
DEBUG: -------------------------------
Ember FastBoot running at http://[::]:3000
```

danmcclain:
Try `FastBoot : 1.0.0-rc.0 (fastboot) / 1.0.0-beta.13 (ember-cli-fastboot)` as a format

danmcclain:
We should only emit this in dev on browser side, and only emit cli-fastboot version in browser outside of dev

index.js
```diff
@@ -140,6 +163,33 @@ module.exports = {
...
+  _getFastBootVersion: function() {
+    var VersionChecker = require('ember-cli-version-checker');
+    var checker = new VersionChecker(this);
+    var fastbootVersionChecker = checker.for('fastboot', 'npm');
+
+    return fastbootVersionChecker.version;
+  },
+
+  _getEmberCliFastBootVersion: function() {
+    var projectPath = process.cwd() + '/node_modules/ember-cli-fastboot';
```
stefanpenner: this isn't the correct path, as `ember s` can be anywhere within the root of your app.

addon/versions.js
```diff
@@ -0,0 +1,6 @@
+import Ember from 'ember';
+
+if (Ember.libraries) {
+  Ember.libraries.register('FastBoot', 'FASTBOOT_VERSION');
```
stefanpenner: if we do this, we should mimic how ED does this:

* ember-data/version module is added
* that module is imported to populate https://github.com/emberjs/data/blob/c9d8212c857ca78218ad98d11621819b38dba98f/addon/-private/core.js#L26

Issue #291, opened 2016-11-14
fastboot.isFastBoot should be false in fallback mode


Redsandro:
I might be wrong because my code sandbox is a bit messy, but when falling back to non-fastboot mode (e.g. on error), it seems that `fastboot.isFastBoot` is still `true`.

This is undesirable, e.g. when using `fastboot.isFastboot` for switching between `fastboot.request.headers` and `window.navigator`. In fallback, the wrong code gets executed because it thinks it's still running in Fastboot.

danmcclain:
What do you mean by fallback mode? Are you referring to when an error happens and the non-FastBoot index.html is served?

rwjblue:
Can you provide a demo repo and instructions to reproduce what you are experiencing?

Redsandro:
> Are you referring to when an error happens and the non-FastBoot index.html is served?

Exactly. If the entire app works but a certain route accidentally has a (fastboot-only) error, fastboot falls back so the app still runs. But `fastboot.isFastboot` is still true.

Or perhaps I am using it wrong? Is there a different flag to check if this is a fastboot-render or normal (client) render?

Redsandro:
@rwjblue 
> Can you provide a demo repo and instructions to reproduce what you are experiencing?

Let me try and just show some code first. Perhaps you can already notice something without me getting my hands dirty on a repo. :wink: 

Because if your initial thought is not: _"Yes, `isFastboot` is always `true` when running a fastboot server, even in fallback mode"_, then I probably made a mistake myself.

```
import Ember from 'ember';

export default Ember.Controller.extend({
	fastboot : Ember.inject.service(),
	init : function() {
		this._super(...arguments);

		Ember.run.schedule("afterRender", this, function() {
			if (!this.get('fastboot').isFastboot)
				this.send("onLayoutComplete");
		});
	},
	actions : {
		onLayoutComplete() {
			Ember.Logger.info("This should only be executed in fallback mode");
			// Some random jQuery
			$('body').css({backgroundColor: 'orange'});
		}
	}
});
```

Fastboot still fails because `$` is (obviously) not defined. I would expect the `isFastboot` switch to prevent this.

danmcclain:
The FastBoot service is destroyed after visit completes, so it is false in the run later. Try wrapping the run later in the FastBoot check, rather than checking inside of the run loop

Redsandro:
[...]

PR #303, opened 2016-12-14
Static isFastBoot prop

jasonmit:
`isFastBoot` doesn't change over the lifetime of the app, doesn't need to be recomputed so doesn't need to be a computed.  Made it also not writable as part of this change.

This is to avoid being trolled #291 

* [x] fix test

/cc @danmcclain

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ cache:
 before_install:
   - "npm config set spin false"
   - "npm install -g npm@^2"
+  - npm install phantomjs-prebuilt
+  - node_modules/phantomjs-prebuilt/bin/phantomjs --version
 
 install:
   - npm install -g bower
--- a/app/services/fastboot.js
+++ b/app/services/fastboot.js
@@ -60,7 +60,7 @@ const Shoebox = Ember.Object.extend({
   }
 });
 
-export default Ember.Service.extend({
+const FastBootService = Ember.Service.extend({
   cookies: deprecatingAlias('request.cookies', { id: 'fastboot.cookies-to-request', until: '0.9.9' }),
   headers: deprecatingAlias('request.headers', { id: 'fastboot.headers-to-request', until: '0.9.9' }),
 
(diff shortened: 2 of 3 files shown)


tomdale:
:+1:, counter-troll measure deployed.

jasonmit:
@tomdale good to go -- just had to bump phantom

bcardarella:
This change makes it difficult to unit test fastboot specific functionality because we can no longer stub this value.

jasonmit:
Could you set the `Fastboot` global?

bcardarella:
yes, we could register a stub service but then depending upon the use case we'd have to stub a whole bunch of other functionality that we'd normally get for free with the Fastboot service

jasonmit:
Sorry, I meant set `Fastboot` before the proto is materialized: https://github.com/jasonmit/ember-cli-fastboot/blob/6302f96e58b005e88f29b5bb0ea8e76def11fca7/app/services/fastboot.js#L101

bcardarella:
@jasonmit I see, isn't that code inferred at load time? If `Fastboot` needs to be `true` for a single unit test it seems to me the opt in/out is all or nothing for the entire runtime.

jasonmit:
> If Fastboot needs to be true for a single unit test it seems to me the opt in/out is all or nothing for the entire runtime.

That is true.  I don't believe I can go back to using `get()` instead of value + writeable because of inconsistencies Phantom (that was my original implementation).  But perhaps that has since been resolved and I just need to bump phantom.

Let me play around with this...

bcardarella:
> would need to be set before the service module is resolved

I wish this were the case, but the AMDs are cached after their first use. So despite the unit tests providing isolation in the way of Ember's own resolver the LOC in question does not get eval'd more than once for the runtime of the test suite.

jasonmit:
I opened a PR to make it writeable again.

bcardarella:
@jasonmit oh nice, thank you :)

PR #312, opened 2016-12-24
remove ember-test-helpers

kellyselden:
appears to be unused


rwjblue:
Thanks for the cleanup, but please try to consolidate PR's next time, it makes things much easier for both the submitter and reviewer.

Issue #270, opened 2016-09-26
Allow loading other assets in fastboot sandbox


kratiahuja:
Currently fastboot only loads the app and vendor files in sandbox via the dist/package.json `manifest` key. Apps may have other assets that may needed to be loaded upfront on the server side. For example, what if the app split the templates into its own bundle or lazy engines. 

We need to allow a configuration to provide other assets to load under the fastboot config in `environment.js`. 

I will work on this issue as part of integrating fastboot infrastructure at LinkedIn.

cc: @tomdale @arjansingh @danmcclain

kratiahuja:
We also need to allow an order of assets to be loaded. For example, we have a usecase where we need to load an asset between vendor and app.

kratiahuja:
This is fixed and merged.

PR #325, opened 2016-12-29
Create an array of app & vendor files for fastboot manifest

kratiahuja:
This PR is to address a [usecase](https://github.com/ember-fastboot/ember-cli-fastboot/issues/270) where an app may want to load other assets after the vendor asset is loaded and before the app asset is loaded in the sandbox. Primarily for example, if we want to load i18n translations file, lazy engine assets. Making the appFiles an array allows other addons that will run after ember-cli-fastboot to update package.json with the order.

It does so by making `appFile` an array in FastBoot manifest. Any addon (running after ember-cli-fastboot) now has a low level primitive to update the `appFiles` array with the order in which they would like load the assets in sandbox.

Also found a bug with fingerprinting assets in FastBoot. We should be using `generateAssetMap` option of `broccoli-asset-rev` to see if assetMap.json is generated.

**Note**: This was discussed a few weeks ago in FastBoot meeting and we decided to provide a low level primitive for advanced addons to override the package.json manifest with the order of assets they would like to load.

Corresponding FastBoot [PR](https://github.com/ember-fastboot/fastboot/pull/105)

TODO:

- [x] Once FastBoot [PR](https://github.com/ember-fastboot/fastboot/pull/105) lands release it and update fastboot-express-middleware
- [x] Fix tests once FastBoot [PR](https://github.com/ember-fastboot/fastboot/pull/105) lands and is released
- [x] Come up with a schema versioning after discussing with @rwjblue

### DIFF
--- a/lib/broccoli/fastboot-build.js
+++ b/lib/broccoli/fastboot-build.js
@@ -24,7 +24,7 @@ function FastBootBuild(options) {
   })[0];
 
   if (assetRev && assetRev.options) {
-    this.assetMapEnabled = !!(assetRev.options.enabled && assetRev.options.assetMapPath);
+    this.assetMapEnabled = !!(assetRev.options.enabled && assetRev.options.generateAssetMap);
 
     if (assetRev.options.assetMapPath) {
       this.assetMapPath = assetRev.options.assetMapPath;
--- a/lib/broccoli/fastboot-config.js
+++ b/lib/broccoli/fastboot-config.js
@@ -5,7 +5,7 @@ var md5Hex = require('md5-hex');
 var path   = require('path');
 var Plugin = require('broccoli-plugin');
 
-var LATEST_SCHEMA_VERSION = 1;
+var LATEST_SCHEMA_VERSION = 2;
 
 function FastBootConfig(inputNode, options) {
   Plugin.call(this, [inputNode], {
@@ -14,6 +14,7 @@ function FastBootConfig(inputNode, options) {
 
   this.project = options.project;
   this.name = options.name;
+  this.assetMapEnabled = options.assetMapEnabled;
   this.ui = options.ui;
   this.fastbootAppConfig = options.fastbootAppConfig;
   this.outputPaths = options.outputPaths;
@@ -121,8 +122,8 @@ FastBootConfig.prototype.buildManifest = function() {
   var vendorFile = 'fastboot/' + vendorFileName + '.js';
 
   var manifest = {
-    appFile: appFile,
-    vendorFile: vendorFile,
+    appFiles: [appFile],
+    vendorFiles: [vendorFile],
     htmlFile: this.htmlFile
   };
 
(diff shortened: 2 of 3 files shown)


kratiahuja:
@rwjblue Yes I think at some point we will need to explicitly call out what version of `ember-cli-fastboot` is supported by `fastboot` (most likely in a range of semvers when we annouce post FastBoot 1.x releases). 

I don't agree we should emit both `appFile` and `appFiles` from the build step since it causes confusion to users which one is really used by `fastboot` or for a matter of fact which one I should override if I am really doing something advance (example lazy engine support, i18n support). `fastboot` can always maintain the backward compatibility without the users seeing such changes in `package.json`. This is  the reason I decided to keep `fastboot` backward compatible (ie it looks for `appFiles` or for `appFile`). 

There is an RFC (by @danmcclain ) to emit out `fastboot` and `ember-cli-fastboot` version in the console ui which we discussed in the last fastboot meeting.

rwjblue:
> I don't agree we should emit both appFile and appFiles from the build step since it causes confusion to users which one is really used by fastboot or for a matter of fact which one I should override if I am really doing something advance

Yep, agreed that it seems bad. I'm most raising the concern as this is the first (AFAICT) breaking change between `ember-cli-fastboot` and `fastboot`. This is compounded by the fact that we are already in the 1.0.0-{beta,rc} stages with these packages and do not have the ability to signal a "breaking change" to our users without jumping to 2.0.0-beta.

> fastboot can always maintain the backward compatibility without the users seeing such changes in package.json.

Right, I totally agree with you here, except that the issue I'm concerned with is *forward* compatibility.  

Take this example: a user uses the latest ember-cli-fastboot (after this lands and is released) in their development environment and since development is decoupled from deployment, they forget (or don't know it is needed) to update their version of `fastboot`.  In this example, the user will get a super obtuse and hard to reason about error in their production environment.

I am definitely not blocking this (I think its a great change!), but I do think we need to figure this out generally speaking.

---

My concrete proposal is to incorporate a `ember-cli-fastboot` / `fastboot` interface version (only incrementing when there is a protocol change between them), release a patch update to `fastboot` that checks for this number and provides a warning when no version is present and a useful error when the version of ember-cli-fastboot that generated the bundle is not compatible with the fastboot version serving it.

kratiahuja:
@rwjblue Agree with you on the forward compatibility. I wasn't aware of that deploying is totally decoupled from development. I agree with your proposal too. Going forward (before FastBoot 1.0 is released) we could have an ability where if there is an incompatible version between ember-cli-fastboot and fastboot we simply error hard on deploy. This way the owners will realize during deploy itself and not during runtime requests that something is broken. 

I think we should discuss this in the next FastBoot meeting with @tomdale and everyone else. I would prefer we come up with a strategy soon in order to not cause pain to our users.

kratiahuja:
This PR is rebased and tests are fixed now.

PR #327, opened 2017-01-08
baseline rootURL test before any changes.

kellyselden:


kellyselden:
This adds a test with rootURL enabled to see how it behaves before making any actual rootURL changes.

PR #334, opened 2017-01-13
add rootURL support to redirects

kellyselden:


jasonmit:
The NoneLocation should have the rootURL set when the router instantiates it:
https://github.com/emberjs/ember.js/blob/v2.10.0/packages/ember-routing/lib/location/none_location.js#L42 (I fixed this a few versions back in Ember)

I would use this over reading from the config as the values could be different.  I know a few apps that dynamically generate the rootURL to support things like locale prefixing. /en/home, /fr/home, etc.

kellyselden:
@jasonmit Updated to great success.

PR #336, opened 2017-01-13
Allow najax errors to bubble up to the error callback

demersus:
We are seeing that errors, caused from najax being unable to parse json in the response, are causing the fastboot server to crash.  This can be reproduced when an api returns an html response when the expectation is json.  Examples:  Gateway Errors, Redirects, 404 pages.

This catches the error, and forwards it to the options.error function which najax does not even use currently.

### DIFF
--- a/app/initializers/fastboot/ajax.js
+++ b/app/initializers/fastboot/ajax.js
@@ -19,7 +19,11 @@ var nodeAjax = function(options) {
   }
 
   if (najax) {
-    najax(options);
+    najax(options).fail(function(jqXHR, statusText, e) {
+      if('error' in options) {
+        options.error(jqXHR, statusText, e);
+      }
+    });
   } else {
     throw new Error('najax does not seem to be defined in your app. Did you override it via `addOrOverrideSandboxGlobals` in the fastboot server?');
   }


demersus:
I have a PR open here that will allow us to extract the status code in the error callback:
https://github.com/najaxjs/najax/pull/53

kratiahuja:
Curious to understand, does the error not bubble upto `Ember.onerror`? If not, IMO the ember-data should to be updated such that `_ajaxRequest` returns a promise.

demersus:
Well, the adapter's `ajax()` method returns a promise. But it uses the error and success callbacks of the jQuery (najax) implementation, instead of the returned promise.   `_ajaxRequest`  simply forwards all the options it receives to the jquery ajax implementation.  It doesn't return any promise.  Unless i'm missing something, errors will bubble up, uncaughtm causing fastboot to crash.  It there a better way to handle the error condition than a promise rejection from the adapter's `ajax()` method?

demersus:
Actually, I think this is a shortcoming of the najax library.  It should be detecting if there is an error callback, and calling it.  Which is how jQuery behaves in the browser.  I'll look into opening a PR there, and put the link here.

PR #338, opened 2017-01-18
process.env.FASTBOOT_DISABLED to disable fastboot build

kellyselden:
This would be a pre 1.0 power user flag to disable the fastboot build while retaining the fastboot service. Use case is large legacy apps that already take forever to build. Developing locally would like the option to disable fastboot while CI does the fastboot builds. I tried blacklisting the addon, but then I got `Error: Attempting to inject an unknown injection: 'service:fastboot'` on CI because the fastboot service was gone.

I brought this up in slack and there was at least one other user with a fork of fastboot doing this check as well as me.


jasonmit:
We took a similar approach, would be nice if this landed :)

tomdale:
I don't love this, but it's definitely solving a real need. Let's keep it around until the double build problem is addressed. @kellyselden, can you make sure we note this flag in the README so it's not just a secret handshake you have to learn about?

kellyselden:
Added to README. @tomdale need your approval on my wordsmithing.

PR #339, opened 2017-01-18
path needs rootURL massaging earlier to pass the isTransitioning check

kellyselden:


tomdale:
This feels like it is working around a bug in Ember. I think we should merge this now to address the issue, but @kellyselden will file a bug on the Ember.js repo to make sure that the form of the path is normalized between initial routes and transitions.

PR #341, opened 2017-01-20
Use npm 3 in travis to solve tests for #269

josemarluedke:
This fixes the tests on PR #269.

cc @rwjblue @danmcclain.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,6 @@ cache:
 
 before_install:
   - npm config set spin false
-  - npm install -g npm@^2
   - npm install -g bower
   - bower --version
   - npm install phantomjs-prebuilt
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -22,7 +22,7 @@ cache:
 
 before_install:
   - npm config set spin false
-  - npm install -g npm@^2
+  - npm install -g npm@^3
```
kellyselden: I think you can remove this line to get npm 3 automatically.

kratiahuja:
Does this mean we are only going to support npm 3? Just want to make sure we call it out explicitly.

josemarluedke:
@kratiahuja Don't think ember-cli-fastboot is only compatible with NPM 2, just that the test suite breaks when using NPM 2. Right now, the test suite is broken for users that are using NPM 3 or above.

.travis.yml
josemarluedke: Will update!

PR #345, opened 2017-02-03
Add hooks to fastboot server task

mfazekas:
Not sure if this is the best way to solve, but we'd like to add route handlers to the express server started by fastboot. One concrete usecase is `http-proxy` to proxy api requests to real backend, other usecase is running mirage as express middleware.

Right now i'm using it by adding `app/instance-initializers/fastboot/start-proxy.js`. In there i do this:

```es6
function proxy(app) {
  var httpProxy = FastBoot.require('http-proxy');
   ...
   var proxy = httpProxy.createProxyServer(proxyOptions);
   ...
   app.use("/api", function(req, res, next) {
      ...
      proxy.web(req, res, {target: proxyTarget});
   })
}

let FastbootServerTask = FastBoot.require('ember-cli-fastboot/lib/tasks/fastboot-server');
FastbootServerTask.on('start', function (app) {
   proxy(app);
});
...
```

### DIFF
--- a/lib/tasks/fastboot-server.js
+++ b/lib/tasks/fastboot-server.js
@@ -8,7 +8,7 @@ const http = require('http');
 const path = require('path');
 const parseStackTrace = require('../utilities/parse-stack-trace');
 
-module.exports = CoreObject.extend({
+let FastbootServerTask = CoreObject.extend({
 
   exec,
   http,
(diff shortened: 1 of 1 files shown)


kratiahuja:
We are in process of unifying the fastboot serve experience with `ember-cli` that fastboot command will no longer be needed. `ember-cli` already provides a proxy option which will work in fastboot (once unified). 

IMO we should hold this off since we will be deprecating fastboot command soon(TM)

mfazekas:
@kratiahuja i agree, that sure that sounds like a much better solution

PR #346, opened 2017-02-03
Fix server never getting started when using --no-build flag

simonihmig:
When running `ember fastboot --no-build` (which is mentioned in the README for a fast debug mode) the server is never started, as it is waiting endlessly for the `postBuild` event, which is never fired when skipping the build step.

This should fix it, immediately calling `start` after `run` when `options.build` is false.


rwjblue:
Thank you!

PR #348, opened 2017-02-05
Bump fastboot-express-middleware version

kratiahuja:
Needed to fix one test. The failing test was because when we there is an error from fastboot and it is not UnrecognizedUrl, we don't close the response but mark it to 500 and call the next middleware. Express then sets the correct response and content type.

cc: @rwjblue


rwjblue:
:+1:

PR #354, opened 2017-02-10
Expose request.method and request.body in Fastboot service

bcardarella:
Related: https://github.com/emberjs/rfcs/pull/185

### DIFF
--- a/app/services/fastboot.js
+++ b/app/services/fastboot.js
@@ -11,6 +11,8 @@ const RequestObject = Ember.Object.extend({
     let request = this.request;
     delete this.request;
 
+    this.method  = request.method;
+    this.body    = request.body;
     this.cookies = request.cookies;
     this.headers = request.headers;
     this.queryParams = request.queryParams;


bcardarella:
There are no tests for any parts of the request object: https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/tests/unit/services/fastboot-test.js

danmcclain:
There are 2 tests folders (test and tests). The one you link to are the client side tests, server request tests are here https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/test/request-details-test.js

In the test you link to, we verify that `fastboot.request` is null on the client side

danmcclain:
Also, this won't work without an additional change to ember-fastboot/fastboot, as we purposely don't expose the Express request object: https://github.com/ember-fastboot/fastboot/blob/master/src/fastboot-request.js

rwjblue:
FWIW, I think that exposing the `request.method` and `request.body` is quite different than implementing all of https://github.com/emberjs/rfcs/pull/185.  I am not certain if there are security implications here, but assuming that we don't have security concerns it seems fine to expose them regardless of if we decide to embrace POST requests...

bcardarella:
@rwjblue exposing `method` and `body` allows you to build a workflow as described in emberjs/rfcs#185

because we can muck with the middleware stack, whereas the RequestObject is private and cannot make any changes

rwjblue:
@bcardarella - Ya, I understand how it can be used.  I'm just saying that exposing them from the request likely seems fine regardless of what we do on that RFC.

tomdale:
@bcardarella Hey Brian, this looks great. If you can add tests we would be happy to merge.

bcardarella:
@tomdale working on that atm. Should be coming in a bit

bcardarella:
https://github.com/ember-fastboot/fastboot/pull/124 needs to be merged before this PR's tests can go green

tomdale:
@bcardarella Just merged that one, we'll cut a release and then re-run these tests. Thanks!

danmcclain:
FastBoot version 1.0.0-rc.4 has been released

bcardarella:
In order to get the integration test for `request.body` passing the `bodyParser` middleware has to be brought in. I'm reluctant to suggest that this is brought into the default stack in `ember-cli-fastboot` as I suspect that will meet resistance and require a deeper examination of this PR. I would like to add an in-repo addon to the generated addonTest app the integration test uses. The addon would inject itself *before* `ember-cli-fastboot` in the addon stack and bring in `bodyParser`.

It would then be required for anyone using this feature to opt-in by bringing `bodyParser` in on their own app rather than getting it for free.

However, based upon what I've seen of `ember-cli-addon-tests` it doesn't seem I can make this happen for a specific test. It will be an all or nothing opt-in for the group of request tests (I think).

Is there any resistance to this approach before I spend time on implementing?

tomdale:
@bcardarella Why do we need to pull in `body-parser`? We should be able to test against the raw, unparsed `body` property, right?

bcardarella:
@tomdale sounds right, my node-fu isn't strong. I've always used body-parser. I'll take a look at just the asserting the raw body. I assume this comes in as a buffer?

bcardarella:
@tomdale I believe the issue with asserting the raw body is that this library doesn't have access to the original request object. It is built in `ember-fastboot`. bodyParser would be necessary higher up the middleware chain to give us the `request.body` property to use here: https://github.com/ember-fastboot/fastboot/blob/master/src/fastboot-request.js#L12

The solution I was working towards was having ember-cli-addon-tests add an in-repo addon that would add a server middleware that added bodyParser just for that test. However, because https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/lib/tasks/fastboot-server.js is a completely different server environment it does not make any attempt to pull in the addon middleware hook so this is not an option.

I was told that someone is working on rewriting the fastboot server to just work from the ember server. Once this is delivered I should be able to complete this PR.

kratiahuja:
@bcardarella 

> I was told that someone is working on rewriting the fastboot server to just work from the ember server. Once this is delivered I should be able to complete this PR.

That is me 😄 . The PR is here: #356

bcardarella:
@kratiahuja awesome!

danmcclain:
Is it possible to pass the raw body into the assembled request object instead of the assumed parsed body? Some may not use bodyParser and want access to the raw body itself. From my understanding, the way this PR is moving forward is that bodyParser is a requirement at this point. Please let me know if that assumption is incorrect

bcardarella:
bodyParser is only a requirement for getting this test to pass. It wouldn't be a run-time dependency for ember-cli-fastboot. My take on this is that the consumers of this feature can opt-in to using bodyParser if they'd like by adding that middle-ware upstream of the fastboot middleware. Or they can use another method. The only unknown for me is if: https://github.com/ember-fastboot/fastboot/blob/master/src/fastboot-request.js#L12 would be a problem if there is another way to retrieve the body. The Fastboot Request that we get access to is just a proxy to the original request object. So if there are more ways to retrieve that body it may be worth exposing that in the proxy object.

bcardarella:
Working on this now that #356 has landed

rwjblue:
Thanks @bcardarella, sorry for the delays

bcardarella:
To complete this PR I believe I need to make the following change to ember-cli:

https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/server/middleware/history-support/index.js#L58

This LOC restricts all inbound requests to just `GET`. I'd like to propose a change of something like:

```js
let methods = options.allowedHttpMethods || ['GET'];

if (hasHTMLHeader && isForBaseURL && methods.indexOf(req.method) > -1) {
```

the config options are bound to the CLI tool, so the default value could be set upstream in https://github.com/ember-cli/ember-cli/blob/master/lib/commands/serve.js and probably https://github.com/ember-cli/ember-cli/blob/master/lib/commands/test.js as well.

@stefanpenner or @rwjblue would this require an RFC for ember-cli? It wouldn't change the default behavior and would be opt-in only.

tomdale:
I appreciate making this scoped to `GET` by default for security reasons, but it does make sense to allow addons to loosen that restriction now that the Express server is pluggable.

bcardarella:
@tomdale so you're suggesting that the addons would opt in? I would going to allow the dev to opt-in by setting a value in `.ember-cli` would that satisfy this concern?

tomdale:
@bcardarella I'm assuming this is opt-in on the FastBoot side too. I'm just saying ideally you opt-in once and don't have to set configuration options in multiple places.

bcardarella:
the limit right now is on ember-cli that disallows non `GET` requests. I haven't looked into other consumers of the fastboot middleware like the app server. It appears to have `GET` hardcoded:

https://github.com/ember-fastboot/fastboot-app-server/blob/master/src/express-http-server.js#L40-L52

In that case, I believe `.ember.cli` option wouldn't be ideal as this needs to work in non-ember-cli environments. With the way ember-fastboot-appserver is implemented I'm not sure what the best solution would be. Suggestions?

cibernox:
This PR is superseded by #457

Issue #274, opened 2016-10-07
unify ember fastboot --watch + ember server


stefanpenner:
``` js
ember fastboot --watch
```

and 

``` js
ember server
```

are essentially the same feature. Or at the very least `ember server` should also provide fastboot, so that regular development is fastboot aware (catch issues early etc).

---

I believe the crux of the problem, is that there is no public API to make `fastboot` and `ember serve` work together easily.

specifically: https://github.com/ember-cli/ember-cli/blob/master/lib/tasks/server/middleware/serve-files/index.js#L22 is a black box, but we most likely want to make "serving" the actually assets intercept-able. 

Now in theory, one could remove the entire middleware, but this would largely require reimplementing the entire middleware only to add a small hook, which ultimately will just cause future grief as the two diverge, or as a third feature wants to also provide additional behavior here.

broccoli-middleware is a module we control: https://github.com/ember-cli/broccoli-middleware and can make easily make changes to. Most likely [here](https://github.com/ember-cli/broccoli-middleware/blob/master/lib/middleware.js#L109-L111) (but more then likely this whole file will need some refactor to make this work.

The design I believe would work is: "I'm about to serve index.html` tell me what I'm actually supposed to do.

The public handler that would be provided via configurable most likely would look very similar to an express middleware (but promise aware);

``` js
// default/existing handler
function indexHandler(path, response, request, next) {
  return fs.readFileSync(...);
}
```

``` js
// the fastboot variant
function fastbootHandler(path, response, request, next) {
  return app.visit(path).then(result => result.html()).then(html => request.send(html));
}
```
### Goals:
- continue to benefit from existing broccoli middleware features (serving the error page when the build breaks for example)
- pluggable "serving" of the index.html file
- promise aware (so we propagate errors nicely)
- allow for streaming responses, so that known parts (html head etc can be flushed early, and more advanced plugins can stream in data after flushing the minimal viable html. (likely "just work" if we support the above API.
### Stuff to do
- actually design pluggable interface for broccoli-middleware
- implement / refactor broccoli-middleware to support the above design
- make sure ember-cli provides a way to 
### Pre-reqs
- explore more scenarios that may need this, if found consider them.
- ??
### Acceptance
- unify serve + fastboot
- minify divergence

kratiahuja:
Update: Now that the RFC linked here is implemented and merged in. I'll look at fixing this next.

kratiahuja:
WIP PR to fix this issue: #356

PR #356, opened 2017-02-15
Unify fastboot serving with ember-cli

kratiahuja:
**TL;DR: fixes https://github.com/ember-fastboot/ember-cli-fastboot/issues/274 . This will unblock one of the FastBoot 1.0 blockers.** 

## Motivation
Currently FastBoot provides users to serve the base page and the assets via the `ember fastboot` command. It primarily spins up an express server that is responsible for serving the request with a fastboot replaced base page and serves other assets from disk. However, this is similar to what `ember cli` provides via `ember serve`. Since we are not the infrastructure provided by ember-cli, development with `ember fastboot` has become very error prone for various reasons. The PR attempts to resolve those development concerns. 

## How it works
ember-cli@2.12.0-beta.1 and above [introduced a public API](https://github.com/ember-cli/rfcs/pull/80) in ember-cli that allows addons to be responsible for serving the base page and/or assets while still be able to consume the other nice features of `ember serve`. It does so by splitting the task of watching the build and serving assets into different middleware. Infrastructures like FastBoot can then hook a middleware between the watcher and server middleware and be responsible for serving the base page request. 

If apps are using `ember-cli@2.12.0-beta.1` and above, they will be able to serve fastboot rendered base page using `ember serve` and continue using other features of `ember serve`. For apps running ember-cli version below 2.12.0-beta.1, they will not be able to use `ember serve` with FastBoot.

With the new serving logic, you can use `curl` with `ember serve`:
`curl 'http://localhost:4200/' -H 'Accept: text/html'` 

This PR does the following:
- [x] Marks `ember-cli-fastboot` addon to run before `broccoli-serve-files` addon. `broccoli-serve-files` is an in-repo addon in ember-cli
- [x] Uses `serverMiddleware` API of `ember-cli` to provide a middleware to use `fastboot-express-middleware` when there is a base request. 
- [x] Added a message when fastboot is serving the base page
- [x] Added deprecation message in `ember fastboot`
- [x] For any other requests other than base page, it forwards the request to the next middleware in the chain.
- [x] Live reloads the FastBoot sandbox with rebuilt app and vendor files whenever there is a rebuild.
- [x] Fixed existing tests
- [x] Added new tests that test the same functionality as `ember fastboot`

## Advantages
With using `ember serve` to have FastBoot serve the initial base page request we get the following features for free:
- Ability to show build errors when the initial page request is made so the development is easier and better
- Not requiring any infrastructure to explicitly build an express app just for serving.
- Ability to have other assets served by `ember cli` as it does for vanilla app
- Ability to use a proxy server for serving any API requests or any other requests using `ember serve --proxy=http://localhost:3000`
- Ability to have live reload in the browser using `ember-cli` live reload functionality.
- Ability to load the built assets from `tmp` directory instead of dist.

## Deprecation of `ember fastboot`
Since this PR does similar to what `ember fastboot --watch --serve-assets` provides, we will be deprecating `ember fastboot` command. 

## Future TODOs
Following are the minor enhancements that I would like to fix on after this PR lands. They do not block this PR.
- [ ] Make FastBoot live reloading much smarter. We do not need to reload the sandbox if there are only CSS changes. Therefore, we could use `fs-tree-diff` to see if any of the app/vendor files have changed on rebuild and only load it sandbox then.
- [ ] Allow developers to provide `sandboxGlobals` and/or custom sandbox class when serving using FastBoot in development. No one seems to be using this so just added a TODO for now. 
- [ ] Provide a way to turn off fastboot serving at runtime without needing to remove `ember-cli-fastboot`

cc: @rwjblue @tomdale @danmcclain @stefanpenner @josemarluedke @arjansingh

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,6 +6,9 @@ var path = require('path');
 var EventEmitter = require('events').EventEmitter;
 var mergeTrees = require('broccoli-merge-trees');
 var VersionChecker = require('ember-cli-version-checker');
+var FastBootExpressMiddleware = require('fastboot-express-middleware');
+var FastBoot = require('fastboot');
+var chalk = require('chalk');
 
 var patchEmberApp     = require('./lib/ext/patch-ember-app');
 var fastbootAppModule = require('./lib/utilities/fastboot-app-module');
@@ -139,12 +142,66 @@ module.exports = {
     return fastbootBuild.toTree();
   },
 
+  serverMiddleware: function(options) {
+    var emberCliVersion = this._getEmberCliVersion();
+    var app = options.app;
+    var options = options.options;
+
+    if (emberCliVersion.satisfies('>= 2.12.0-beta.1')) {
+      // only run the middleware when ember-cli version for app is above 2.12.0-beta.1 since
+      // that version contains API to hook fastboot into ember-cli
+
+      app.use((req, resp, next) => {
+        var broccoliHeader = req.headers['x-broccoli'];
+        var outputPath = broccoliHeader['outputPath'];
+
+        if (broccoliHeader['url'] === req.serveUrl) {
+          // if it is a base page request, then have fastboot serve the base page
+          // TODO(future): provide a way to turn this off without needing to uninstall this addon
+          if (!this.fastboot) {
+            // TODO(future): make this configurable for allowing apps to pass sandboxGlobals
+            // and custom sandbox class
+            this.ui.writeLine(chalk.green('App is being served by FastBoot'));
+            this.fastboot = new FastBoot({
+              distPath: outputPath
+            });
+          }
+
+          var fastbootMiddleware = FastBootExpressMiddleware({
+            fastboot: this.fastboot
+          });
+
+          fastbootMiddleware(req, resp, next);
+        } else {
+          // forward the request to the next middleware (example other assets, proxy etc)
+          next();
+        }
+      })
+    }
+  },
+
   outputReady: function() {
     this.emit('outputReady');
   },
 
-  postBuild: function() {
+  postBuild: function(result) {
     this.emit('postBuild');
+    if (this.fastboot) {
+      // should we reload fastboot if there are only css changes? Seems it maynot be needed.
+      // TODO(future): we can do a smarter reload here by running fs-tree-diff on files loaded
+      // in sandbox.
+      this.ui.writeLine(chalk.blue('Reloading FastBoot...'));
+      this.fastboot.reload({
+        distPath: result.directory
+      });
+    }
+  },
+
+  _getEmberCliVersion: function() {
+    var VersionChecker = require('ember-cli-version-checker');
+    var checker = new VersionChecker(this);
+
+    return checker.for('ember-cli', 'npm');
   },
 
   _getEmberVersion: function() {
--- a/lib/commands/fastboot-build.js
+++ b/lib/commands/fastboot-build.js
@@ -20,7 +20,7 @@ module.exports = {
       project: this.project
     });
 
-    deprecate("Use of ember fastboot:build is deprecated. Please use ember build instead.");
+    deprecate('Use of ember fastboot:build is deprecated. Please use ember build instead.');
 
     return buildTask.run(options);
   },
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ ember install ember-cli-fastboot
 * `ember fastboot --serve-assets`
 * Visit your app at `http://localhost:3000`.
 
+**Note**: If your app is running ember-cli v2.12.0-beta.1+, you can just use `ember serve` instead of `ember fastboot --serve-assets`.
+
 You may be shocked to learn that minified code runs faster in Node than
 non-minified code, so you will probably want to run the production
 environment build for anything "serious."
@@ -381,7 +383,7 @@ present.
 
 ### Prototype extensions
 
-Prototype extensions do not currently work across node "realms."  Fastboot 
+Prototype extensions do not currently work across node "realms."  Fastboot
 applications operate in two realms, a normal node environment and a [virtual machine](https://nodejs.org/api/vm.html).  Passing objects that originated from the normal realm will not contain the extension methods
 inside of the sandbox environment. For this reason, it's encouraged to [disable prototype extensions](https://guides.emberjs.com/v2.4.0/configuring-ember/disabling-prototype-extensions/).
 
--- a/lib/commands/fastboot.js
+++ b/lib/commands/fastboot.js
@@ -4,6 +4,7 @@ const RSVP = require('rsvp');
 const getPort = RSVP.denodeify(require('portfinder').getPort);
 const ServerTask = require('../tasks/fastboot-server');
 const SilentError = require('silent-error');
+const VersionChecker = require('ember-cli-version-checker');
 
 const blockForever = () => (new RSVP.Promise(() => {}));
 
@@ -28,12 +29,14 @@ module.exports = function(addon) {
     ServerTask,
 
     run(options) {
+      const printDeprecations = () => this.printDeprecations(options);
       const runBuild = () => this.runBuild(options);
       const runServer = () => this.runServer(options);
       const startServer = (serverTask) => this.startServer(serverTask, options);
       const blockForever = this.blockForever;
 
       return this.checkPort(options)
+        .then(printDeprecations)
         .then(runServer) // starts on postBuild SIGHUP
         .then(options.build ? runBuild : startServer)
         .then(blockForever);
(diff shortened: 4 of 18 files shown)


kratiahuja:
This is now ready for review.

kratiahuja:
Note to self:
- [ ] Test an app with containing broccoli-asset-rev and make sure it still uses fingerprinted assets in manifest.

index.js
```diff
@@ -139,12 +142,65 @@ module.exports = {
...
 
-  postBuild: function() {
+  postBuild: function(result) {
     this.emit('postBuild');
+    if (this.fastboot) {
+      // should we reload fastboot if there are only css changes? Seems it maynot be needed.
+      // TODO: we can do a smarter reload here by running fs-tree-diff on files loaded
+      // in sandbox (explore later).
+      this.ui.writeLine('Reloading FastBoot...');
+      this.fastboot.reload({
```
stefanpenner: Is fastboot `reload` synchronous? If it is asynchronous, we may need to do more work to ensure no requests during this period are served.

index.js
```diff
@@ -139,12 +142,65 @@ module.exports = {
...
+            this.fastboot = new FastBoot({
+              distPath: outputPath
+            });
+          }
+
+          var fastbootMiddleware = FastBootExpressMiddleware({
+            fastboot: this.fastboot
+          });
+
+          fastbootMiddleware(req, resp, next);
```
stefanpenner: the conditional on line 160 suggests we only serve `fastbootMiddelware` for the first request, subsequent requests will fallback to the `else` case.

kratiahuja: Yes that is correct. If I understand correctly, FastBoot is only meant for initial requests and that too for the base page request (basically have your app be server side rendered for the initial request). For all other requests (other assets, subsequent transitions) should be defaulted to `ember-cli`'s asset serving. 

Do you think that is a wrong assumption? Is there a usecase that I might have missed?

index.js
rwjblue: We can implement the `instrumentation` API (ember-cli@2.13+) and detect the output files that have changed, to see if we need to reload `fastboot` here.

kratiahuja: Agreed. That will eliminate a lot of boiler plate code that I planned to write :)

josemarluedke:
Awesome work here @kratiahuja! I just tried this branch in our application and I got the following error:

```
$ ember s
Livereload server on http://localhost:49153
Serving on http://localhost:4200/

Build successful - 18687ms.



Slowest Nodes (totalTime => 5% )              | Total (avg)
----------------------------------------------+---------------------
EslintValidationFilter (5)                    | 4874ms (974 ms)
TestTransformFilter (2)                       | 3551ms (1775 ms)
Babel (88)                                    | 3349ms (38 ms)
SassCompiler (3)                              | 1745ms (581 ms)
Concat (16)                                   | 1292ms (80 ms)
Funnel (248)                                  | 1244ms (5 ms)

App is being served by FastBoot
TypeError: Path must be a string. Received undefined
    at assertPath (path.js:7:11)
    at Object.join (path.js:1211:7)
    at EmberApp.readPackageJSON (/Users/josemarluedke/Projects/main/node_modules/fastboot/dist/cjs/ember-app.js:309:27)
    at new EmberApp (/Users/josemarluedke/Projects/main/node_modules/fastboot/dist/cjs/ember-app.js:29:27)
    at FastBoot._buildEmberApp (/Users/josemarluedke/Projects/main/node_modules/fastboot/dist/cjs/index.js:98:21)
    at new FastBoot (/Users/josemarluedke/Projects/main/node_modules/fastboot/dist/cjs/index.js:47:14)
    at Class.<anonymous> (/Users/josemarluedke/Projects/main/node_modules/ember-cli-fastboot/index.js:164:29)
    at Layer.handle [as handle_request] (/Users/josemarluedke/Projects/main/node_modules/express/lib/router/layer.js:95:5)
    at trim_prefix (/Users/josemarluedke/Projects/main/node_modules/express/lib/router/index.js:312:13)
    at /Users/josemarluedke/Projects/main/node_modules/express/lib/router/index.js:280:7
GET / 500 5.984 ms - -
GET /favicon.ico 200 1.707 ms - -
```

```
$ ember -v
ember-cli: 2.12.0-beta.1
node: 7.5.0
os: darwin x64
```

mfazekas:
@kratiahuja with this change `ember serve` will be serving fastboot version if i point my browser to http://localhost:4200/. Is there a way to ask for regular non fastboot other than remove `ember-cli-fastboot` from package.json? Usecase is that you see some error with fastboot and want to check if it happens in non fastboot version too.

danmcclain:
Just spitballing: what if we could skip injected middleware via query params

kellyselden:
@mfazekas @danmcclain https://github.com/ember-fastboot/ember-cli-fastboot#double-build-times-and-no-incremental-builds this is how you can run a normal build without removing the ember-cli-fastboot package.

danmcclain:
@kellyselden I was thinking of a way that wouldn't require you to restart the server (and also double build times are going away 🙌 )

index.js
```diff
@@ -139,12 +142,65 @@ module.exports = {
...
+        } else {
+          // forward the request to the next middleware (example other assets, proxy etc)
+          next();
+        }
+      }.bind(this))
```
kratiahuja: Yeah I totally forgot ember-cli allows arrow functions now. Will update this.

kratiahuja:
@mfazekas @danmcclain I agree this is a valid usecase but not scoped to this PR. I want to first get base functionality in and then work on this enhancement. I also like the idea of the query param to switch fastboot serving on/off at run time. I was also thinking of providing it via `.embercli` config file if possible. Mind opening a separate issue for this and I will be happy to address it once this is merged? I would like to evaluate all options on the table and implement the best feasible one.

index.js
kratiahuja: Updated to use arrow functions.

index.js
stefanpenner: > Do you think that is a wrong assumption? Is there a usecase that I might have missed?

Ya, that isn't quite correct. If a user hard reloads, they should once again get a fastboot rendered page.

kratiahuja: @stefanpenner They will get a fastboot rendered page. FastBoot is just responsible for serving the index.html with the rendered content. I did test both use cases on a test app.  I am not sure I am following what is not correct.

stefanpenner: Your right. for some reason i read line 174 being within the if statement of 161. Not sure how that happened, my bad.

kratiahuja:
Tested this PR with `fastboot-website` app on ember-cli version `2.12.0-beta.1` and `2.11.0` (this is the current one) and both worked as expected with `ember fastboot --serve-assets` and `ember serve` commands.

mfazekas:
We also tested with our app [percy-web](https://github.com/percy/percy-web/pull/69) and it was working great, we needed it for proxy support. (Only issue is `FASTBOOT_DISABLED` is broken with this PR as it disables fb build but this PR tries to serve it, but that should be addressed in #358).

kratiahuja:
@rwjblue could you please review this? 😄

josemarluedke:
Awesome work @kratiahuja! 

I tested this branch in our application as well a brand new ember app and it works as a charm. 🎉 

Looking forward to have this merged and available to users.

mirague:
This looks good!

bcardarella:
Is there anything else blocking this PR?

kratiahuja:
@bcardarella Nope nothing is blocking this PR except someone reviewing it. I'll ping @rwjblue or someone else again after the ember conference.

bcardarella:
Didn't @stefanpenner already approve?

kratiahuja:
@bcardarella Yes but I think @rwjblue also wanted to review before merging. Unfortunately I don't have merge powers so we will have to wait for someone with merge powers to review and merge.

kratiahuja:
Bumping this again. @rwjblue could you please review?

rwjblue:
Only two super minor changes around messaging, then this is good to go...

lib/commands/fastboot-build.js
```diff
@@ -20,7 +20,8 @@ module.exports = {
       project: this.project
     });
 
-    deprecate("Use of ember fastboot:build is deprecated. Please use ember build instead.");
+    deprecate('ember fastboot:build will be deprecated and removed after FastBoot 1.0 is released.');
```
kratiahuja: Yeah but I just wanted to update the deprecation to say we will be dropping this command too.

kratiahuja:
@rwjblue This is good to merge now. Travis is 💚

rwjblue:
Thank you for all your hard work on this @kratiahuja!

simonihmig:
Awesome, thanks indeed! 🎉

Issue #360, opened 2017-03-07
Master punch list for resolving Fastboot 1.0 build issues


kratiahuja:
This issue with track the things we need to do in order to resolve the fastboot build. This issue supersedes #264, #246 etc. Once this master list is resolved, fastboot build and serve issues should be resolved.

## Pre requisites
With the help of `ember-cli` team, we have come up with a viable solution to resolve the double builds. The solution is described in the [meeting notes here](https://gist.github.com/kratiahuja/fd073007e10abb9db0a2ec42bc1d7c17). 

## Build blockers
Following are the things we need to do in `ember-cli-fastboot`. The sub items under each section are more of a checklist:
- [x] Build fastboot overrides using `treeForPublic` hook per the new build spec. It will works as follows:
   - [x] Collect the `fastboot-app` tree from the project addons and create a funnel for every addon tree with destDir as `appName-fastboot`
   - [x] Collect the `fastboot-app` tree from the root app and create a funnel for the app fastboot tree with destDir as `appName-fastboot`
   - [x] Merge the trees collected from the addons and app and create a new Funnel
   - [x] Run the JS transpilation on the new funneled tree using `ember-cli-preprocess-registry/preprocessors` API. Use the `preprocessJs` method from the preprocessors with providing the registry 
   - [x] Concat the output of transpilation with `outputFile` set to `assets/appName-fastboot.js`
   - [x] Merge the above output tree with the tree returned from `treeForPublic`
- [x] Make sure the `fastboot-app` tree in app is a watched directory tree. Most likely it will be otherwise follow `treeForGenerator` does [here](https://github.com/ember-cli/ember-cli/blob/d209b88581bd2a723ab6e0fbac2f3b7f00a436b7/lib/models/addon.js#L385)
- [x] Update `FastBootConfig` broccoli plugin to add `assets/appName-fastboot.js` in `appFiles` manifest list of package.json. The corresponding code is [here](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/lib/broccoli/fastboot-config.js#L118)
- [x] Update existing intializers of `ember-cli-fastboot` per the new spec
- [x] Expose a hook to add to manifest files
- [x] Remove the old fastboot double build spec
[...]

Issue #292, opened 2016-11-21
EmberApp requires defaults as first argument, otherwise: JavaScript heap out of memory


backspace:
Hello! I’m excited about this. I did some preliminary investigation on how much work it would be to get `travis-web` FastBoot-ing and I quickly ran into an error I didn’t know how to move beyond: `JavaScript heap out of memory`. I’ve put the error at the bottom due to weird GFM problems.

Thanks to @danmcclain and @stefanpenner in the Ember Community Slack, I was able to get past this with [this commit](https://github.com/travis-ci/travis-web/commit/f1302c8e36c4ba2b2f48ba4bc72b57985924b92f) to add `defaults` as the first argument for `EmberApp`.

I see that this is now [the default](https://github.com/ember-cli/ember-cli/blob/7a60e70356e9af566876905bb271bf568e43e687/blueprints/app/files/ember-cli-build.js) in the `ember-cli-build.js` blueprint, so this won’t be a problem for most people. But it could be something others run into when working on this with older Ember CLI applications. @danmcclain mentioned that @arjansingh had this problem too.

Maybe it’s not worth documenting this at all, as I’m not sure how rare this kind of thing is, but at least this issue may help others who are searching for how to handle this.

<details>
<summary>Stack trace from <code>ember fastboot --serve-assets</code></summary>
<pre>
<--- Last few GCs --->

  128427 ms: Mark-sweep 1318.9 (1404.0) -> 1318.9 (1405.0) MB, 405.0 / 0.0 ms [allocation failure] [GC in old space requested].
  128891 ms: Mark-sweep 1318.9 (1405.0) -> 1318.8 (1405.0) MB, 463.7 / 0.0 ms [allocation failure] [GC in old space requested].
  129587 ms: Mark-sweep 1318.8 (1405.0) -> 1326.1 (1404.0) MB, 695.6 / 0.0 ms [last resort gc].
  130250 ms: Mark-sweep 1326.1 (1404.0) -> 1333.7 (1404.0) MB, 662.9 / 0.0 ms [last resort gc].

<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 0x2343144cfb51 <JS Object>
[...]

PR #288, opened 2016-11-10
[WIP]: Prepare fastboot addon for Fasboot 1.0 release

kratiahuja:
In order to unblock Fastboot 1.0 release, we want to make the fastboot build performant. We will achieve this by creating additive assets (see this [issue](https://github.com/ember-cli/ember-cli/issues/6350) for details). This PR primarily addresses some of the work already. 

This PR primarily addresses the following issues:
- [x] Double fastboot builds [issue](https://github.com/ember-fastboot/ember-cli-fastboot/issues/264). Corresponding ember-cli changes are in this [PR] (https://github.com/ember-cli/ember-cli/pull/6395) and [RFC](https://github.com/ember-cli/rfcs/pull/77)
- [x] Removes the monkey patching introduced with fastboot build. Related [issue](https://github.com/ember-fastboot/ember-cli-fastboot/issues/276)
- [x] Updates generated package.json to take in an array of app files (containing app.js and app-fastboot.js). Addresses this [issue](https://github.com/ember-fastboot/ember-cli-fastboot/issues/270)
- [x] Addresses the compatibility story around how to keep this a backward compatible change.
- [x] Corresponding fastboot lib changes are [here](https://github.com/ember-fastboot/fastboot/pull/97)
- [x] Fixes #292

TODO:
- [ ] Cleanup and tidy this work
- [ ] Add more comments to explain what this PR is doing
- [ ] Make fastboot app work using `ember-serve`. Corresponding [issue](https://github.com/ember-fastboot/ember-cli-fastboot/issues/274).
- [ ] Fix tests once ember-cli changes land and this [issue](https://github.com/ember-fastboot/ember-cli-fastboot/issues/274) is fixed.
- [ ] Update documentation
- [ ] Kill `ember fastboot` command
    
cc: @tomdale @stefanpenner @rwjblue @kellyselden @ryanone @danmcclain @arjansingh @ryanone @masonwan @kristoferbaxter

Issue #264, opened 2016-09-15
Impacts Build performance negatively


stefanpenner:
This has been brought up in chat, but i can't seem to find an issue. If there is one, and I missed it feel free to close. 

More concretely it appears this add-on at-least duplicates the build, causing ember-twiddle for example to for example browserify (and everything else) twice.

Specifically these area is one of the offenders:
- https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L130-L139
- https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L120-L125

danmcclain:
Related issue: https://github.com/ember-fastboot/ember-cli-fastboot/issues/159

stefanpenner:
Currently, it builds the whole app twice. Once for browser and once for Fastboot. This has many issues, the most obvious is developer ergonomics (doubling our build costs, isn't acceptable) the less obvious issues are different builds === different bugs, harder to debug etc.etc, another permutation here makes all future efforts that build on this also more complicated.

The problem specifically is:

https://github.com/ember-fastboot/ember-cli-fastboot/blob/bb8c4bfc1c9cd7defed1997c48e75616f91657dd/index.js#L121

* tree: is the already built app
* fastbootTree: is a second build (of nearly the same app) + fastbootisms

The majority of the system is the same between both worlds. With the exception of some custom fastboot replacements/configuration.

Current these faceboot replacements are very coupled to the app, but this does not need to be. If decoupled, we will most likely have simpler and more performance system.

If instead we did the following:

1. Build the app.js like we do today (maybe minor tweaks)
2. Build app-fastboot.js as an additional set of AMD modules that adds new or replaces existing modules 
3. The node vm running fastboot, merely loads (app.js) then loads (app-fastboot.js) and then finally starts the app, the AMD modules will merge as expected.

Anything that prevents this from working, we should identify and come up with targeted solutions.

For example. app/fastboot and having fastboot initializers mixed into app, unnecessarily complicates the build. Instead moving `fastboot` to a sibling of app, allowing it to have its own pipeline (it is going to run in node after all)...

That pipeline for fast-boot stuff, could be something relatively simple:

```
var Concat = require('broccoli-concat');
var Funnel  = require('broccoli-funnel');
var Babel = require('broccoli-babel-transpiler');
var amdNameResolver = require('amd-name-resolver').moduleResolve;

var fastboot = new Funnel('fastboot');
[...]

PR #369, opened 2017-04-02
Refactor fastboot build to be more performant

kratiahuja:
**TL;DR: This unblocks FastBoot 1.0 release with a breaking change.  Resolves most of the issues mentioned in #360, #292, #264, #246   .**

***Sending out a PR since substantial work is done that it can be reviewed for directional purposes.***

**NOTE**: I took a slightly different approach in how the fastboot initializers will be written per the [spec](https://gist.github.com/kratiahuja/fd073007e10abb9db0a2ec42bc1d7c17)

## Fixes in this PR currently are
- [x] Fixes fingerprinted assets to be properly loaded in FastBoot sandbox
- [x] Build fastboot overrides using `treeForPublic` hook per the new build spec. It will works as follows:
   - [x] Collect the `fastboot-app` tree from the project addons and create a funnel for every addon tree with destDir as `appName-fastboot`
   - [x] Collect the `fastboot-app` tree from the root app and create a funnel for the app fastboot tree with destDir as `appName-fastboot`
   - [x] Merge the trees collected from the addons and app and create a new Funnel
   - [x] Run the JS transpilation on the new funneled tree using `ember-cli-preprocess-registry/preprocessors` API. Use the `preprocessJs` method from the preprocessors with providing the registry 
   - [x] Concat the output of transpilation with `outputFile` set to `assets/appName-fastboot.js`
   - [x] Merge the above output tree with the tree returned from `treeForPublic`
- [x] Make sure the `fastboot-app` tree in app is a watched directory tree. Most likely it will be otherwise follow `treeForGenerator` does [here](https://github.com/ember-cli/ember-cli/blob/d209b88581bd2a723ab6e0fbac2f3b7f00a436b7/lib/models/addon.js#L385)
- [x] Update `FastBootConfig` broccoli plugin to add `assets/appName-fastboot.js` in `appFiles` manifest list of package.json. The corresponding code is [here](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/lib/broccoli/fastboot-config.js#L118)
- [x] Update existing intializers of `ember-cli-fastboot` per the new spec
- [x] Remove the old fastboot double build spec
   - [x] Remove usage of private `ember-cli` patching from `ember-cli-fastboot`. See example [here](https://github.com/ember-fastboot/ember-cli-fastboot/pull/288/files#diff-168726dbe96b3ce427e7fedce31bb0bcL55)
   - [x] Remove invoking `fastboot-filter-initializers` in `ember-cli-fastboot`. See relevant example changes [here](https://github.com/ember-fastboot/ember-cli-fastboot/pull/288/files#diff-168726dbe96b3ce427e7fedce31bb0bcL108)
   - [x] Remove `config` function from `ember-cli-fastboot/index.js`
   - [x] Set `app-boot` contentFor correctly. See example [here](https://github.com/ember-fastboot/ember-cli-fastboot/pull/288/files#diff-168726dbe96b3ce427e7fedce31bb0bcL79)
   - [x] Fix `config-module` contentFor hook to not rely on `__is_building_fastboot__` flag. See example fix [here](https://github.com/ember-fastboot/ember-cli-fastboot/pull/288/files#diff-168726dbe96b3ce427e7fedce31bb0bcR133)
   - [x] Remove `preconcat` function as it will be unused in new build spec
   - [x] In `postprocessTree` hook, we should invoke `FastBootConfig` broccoli plugin directly instead of invoking `FastBootBuild` broccoli plugin. See example changes [here](https://github.com/ember-fastboot/ember-cli-fastboot/pull/288/files#diff-168726dbe96b3ce427e7fedce31bb0bcR166)
   - [x] Remove `FastBootBuild` broccoli plugin as it will be unused
- [x] Update tests
- [x] Kill usage of process.env.EMBER_CLI_FASTBOOT flag as it will be unavailable in new build spec
- [x] Expose a hook to add to manifest files
- [x] Migration path for apps relying on `fastboot-filter-initializers` or `app/(instance-)?initializers/fastboot`
- [x] Throw build error if addon has `app/[instance-]initializers/browser/*.js`
- [x] Create a hook for addons to add additional things in fastboot-asset (example `app-lt-2.9.0`) --> `treeForFastBoot`
- [x] Clean/reorganize code
- [x] Expose `process.env.FASTBOOT_NEW_BUILD` so that addons like `ember-cli-head` can support both versions.

## TODO (In subsquent PRs):
- [ ] Unskip tests using `ember-cli-head` once that addon embraces new format.
- [ ] Remove TODOs from `index.js`
- [ ] README updates
- [ ] Doc updates
- [ ] Kill ember fastboot command 🔪 
- [ ] Update some of the exisiting addons to not rely on process.env.EMBER_CLI_FASTBOOT flag or not contain `app/[instance-]initializers/[browser|fastboot]/*.js` folder structure (owner: @simonihmig  since he has already started this work)

cc: @rwjblue @tomdale @danmcclain @ryanone @simonihmig @arjansingh @stefanpenner


kratiahuja:
Please feel free to try this out without any other fastboot complaint addons.

kratiahuja:
#374 resolves `Expose a hook to add to manifest files` TODO

lib/build-utilities/migrate-initializers.js
```diff
@@ -0,0 +1,127 @@
...
+ * given root path and throw an error
+ *
+ * @param {String} rootPath
+ */
+function _checkBrowserInitializers(rootPath) {
+  const isBrowserInitializersPresent = _checkInitializerTypeExists(rootPath, 'browser');
+
+  if (isBrowserInitializersPresent) {
+    var errorMsg = `FastBoot build no longer supports ${rootPath}/app/(instance-)?initializers/browser structure. ` +
+                         `Please refer to www.ember-fastboot.com for a migration path.`;
```
simonihmig: @kratiahuja I have picked up migrating an addon as we have discussed on Slack, and after updating `ember-cli-fastboot` to the latest version of your PR, I immediately ran into this error message, being caused by some other (unknown) addon. 

The problem I see here is that a) probably every existing app will run into this error once trying to migrate to FastBoot 1.0, and b) it gives no clue where that offending initializer actually comes from, which could be quite frustrating. It could be even some nested addon dependency the user does not see in his `package.json`. 

I think it would be cool (in this PR or a follow up) to have some better tracing available, like which addon is causing this, or at least what the initializers actual name is, to be able to effectively investigate into this!

simonihmig: Please forget it! ;) I was not paying enough attention to see it actually does exactly this. 🙄  
Need some more sleep probably...

kratiahuja: Yeah we can't auto-migrate browser based initializers, hence we will need to fail the build. I am happy to refactor the error messages to be more better. Probably sometime this week i'll come up with a migration cheat sheet so that it is easier for addons 😄

simonihmig: That would be awesome! Would be nice to have a link in the above mentioned error message, maybe a dedicated migration page on www.ember-fastboot.com?

lib/build-utilities/migrate-initializers.js
```diff
@@ -0,0 +1,127 @@
...
+  _moveFastBootInitializers(project, hostAppPath);
+}
+
+/**
+ * Function that migrates all addons and host app fastboot initializers to fastboot/app.
+ * It also throws an error if any addon or host app has browser forked initializers.
+ *
+ * @param {Object} project
+ */
+function migrateInitializers(project) {
```
simonihmig: I just noticed that this works on the original files. So in my app after running `ember s` the files were modifed, the old `instance-initializer/fastboot/foo.js` was physically moved to `fastboot/instance-initializer/foo.js` (from POV of git a new, not yet added file). The same for addons, some `yarn link`ed addons got silently modified (which I did not want as that would have broken them for pre 1.0 FastBoot).

I wonder if that migration magic should not happen inside some broccoli tree rather than operating on the original files of the app and in `node_modules/someAddon`?

lib/build-utilities/migrate-initializers.js
```diff
@@ -0,0 +1,127 @@
+'use strict';
+
+const path = require('path');
+const existsSync = require('exists-sync');
+const fs = require('fs-extra');
+
+const fastbootInitializerTypes = [ 'initializers', 'instance-initializers'];
+const FASTBOOT_DIR = 'fastboot';
```
simonihmig: Given the description of the PR I was expecting the new FastBoot folder to be named `fastboot-app` rather than `fastboot`, but this might be just some misunderstanding?

lib/build-utilities/migrate-initializers.js
kratiahuja: Why will it break for pre fastboot 1.0? This migration will not apply for pre fastboot 1.0 but only after.

Reason for moving is because if we don't, these initializers will end up running in browser too. Hence it needs to be moved. I am hoping reading the message, addon owners would move the fastboot intiializers themselves post fastboot 1.0. This migration would be dropped after a few releases of fastboot 1.0

lib/build-utilities/migrate-initializers.js
kratiahuja: I prefer calling it fastboot instead of fastboot-app since it is more generic

lib/build-utilities/migrate-initializers.js
simonihmig: The breaking pre FastBoot 1.0 thing was related to the changes in [ember-useragent](https://github.com/willviles/ember-useragent), that you also reviewed. There we left the `instance-initializer/fastboot/ember-useragent` in place, so it works in pre 1.0 as well as in post 1.0 (because of the automatic migration). But because it was symlinked, and the migration works on the original files, it modified the original addon code, moving the initializer to a the new place...

So this is the point I was trying to make, that this migration did not only move files in a temporary state while building the app like a broccoli plugin would do (take the previous approach of `fastboot-filter-initializers` as an example, which filters one of the `browser|fastboot` initializers in a broccoli tree), but it modified the *original* in-repo files of the app, as well as any symlinked addon (in fact any addon in node_modules). I would expect something like a codemod to do that, but certainly not `ember s`.

kratiahuja: I am fine with `codemod` doing something like this. But what if they forget to run the codemod tool post 1.0 before serving? It will break for them more. I guess we can detect if they ran codemod too and not serve.

simonihmig: I am actually not arguing in favor of a codemod, I am just arguing that `ember s` should not modify any original file in my app/addons. It can move things around in a broccoli tree when building, but not the original in-repo files.

kratiahuja: @simonihmig wouldn't moving in broccoli-tree be very expensive? Are you talking about after the `treeForApp` is generated? I don't think we will solve the build performance problem then (at least in a temporary sense). 

I totally agree it is not the best way and I am happy to explore new ways to do it. Do you have an example that you can explain with?

Reason for changing the original in-repo files is (maybe moving in broccoli tree solves the below issue):
the problem not moving in-repo files is (as I mentioned earlier) that if we don't move you will end up with two AMD modules after the build:
1. `<app-name>/initializers/fastboot/x.js` <-- This one will get loaded in browser and FastBoot
2. `<app-name>/initializers/x.js` <--- this one is only loaded in FastBoot

When the app will boot in the browser, it _will_ run `app/initializers/fastboot/x.js` (this because we no longer filter the fastboot initializers) which will cause issues. Hence there is a need to _move_ the in-repo files so that we don't generate `<app-name>/initializers/fastboot/x.js`. 

Maybe there is a better way to solve this problem that I am not seeing?

lib/build-utilities/migrate-initializers.js
kratiahuja: Sorry I missed replying to this. Yes would be nice to have a dedicated migration page. I think before this PR is merged, we should send in a warning (but getting another PR) and link to the migration page.

kratiahuja:
Draft and WIP cheatsheet on how  fastboot compatible addons can continue to be backward compatible (with this build changes) if they do these changes: https://gist.github.com/kratiahuja/d22de0fb1660cf0ef58f07a6bcbf1a1c 

This may change subject to the feedback on this PR and feedback on the cheatsheet. I'll try bringing this up in the next FastBoot meeting.

cc: @simonihmig

kratiahuja:
Removing WIP since this is ready to review. The review has got pretty large that I would like to do the remaining TODOs in a subsquent PR.

index.js
```diff
@@ -79,67 +87,133 @@ module.exports = {
...
+      // walk through each addon and grab its fastboot tree
+      var currentAddonFastbootPath = path.join(nodeModulesPath, addon.name, 'fastboot');
+      if (existsSync(currentAddonFastbootPath)) {
+        var fastbootTree = fastbootBuilder.addFastBootPath(currentAddonFastbootPath);
+        fastbootTrees.push(fastbootTree);
+      }
+
+      // invoke addToFastBootTree for every addon
+      if (addon.addToFastBootTree) {
+        var additionalFastBootTree = addon.addToFastBootTree();
```
simonihmig: Related to this problem: https://github.com/ronco/ember-cli-head/pull/21#issuecomment-297839459
Is it actually required that addons return a tree that was created by `lib/build-utilities/fastboot-builder.js`? Isn't it possible that the addon just return a simple tree (like `this.treeGenerator(path)`), and the `mergedFastBootTree` (Line 152) is then moved to `appName` (Funnel destDir)?

kratiahuja: @simonihmig Fixed. You can just use `this.treeGenerator(path)` and return that.

stefanpenner:
Left some thoughts, going to test it out manually next :)

kratiahuja:
@stefanpenner Thank you so much for taking time to review a very very long PR :) Really appreciate it. I'll address the comments this weekend!

test/head-content-test.js
```diff
@@ -4,7 +4,8 @@ var request = RSVP.denodeify(require('request'));
 
 var AddonTestApp = require('ember-cli-addon-tests').AddonTestApp;
 
-describe('head content acceptance', function() {
+// TODO: Unskip this test once ember-cli-head is released with new fastboot build spec.
+describe.skip('head content acceptance', function() {
```
kratiahuja: Yes :)

stefanpenner:
> I'll address the comments this weekend!

Some of them are intended more as discussion (don't assume I'm right :P)

kratiahuja:
> Some of them are intended more as discussion (don't assume I'm right)

Yup I meant I'll reply back this weekend :)

stefanpenner:
Local testing is looking good, rebuild times feel great. Everyone seems to work.

I was a tad surprised that `curl 'http://localhost:4200/'` didn't include the fast boot content but
 `curl 'http://localhost:4200/' -H 'Accept: text/html'` did

index.js
kratiahuja: @stefanpenner I actually like the idea of having `treeForFastBoot` which is more aligned to other existing APIs. And yes it would cover the usecase I wanted it to cover as well. Will update it to providing `treeForFastBoot` :)

index.js
```diff
@@ -2,66 +2,74 @@
...
-
 module.exports = {
   name: 'ember-cli-fastboot',
 
+
   init() {
     this._super.init && this._super.init.apply(this, arguments);
+    // set a environment variable to allow addons to use `fastboot-filter-initializers`
+    // for old versions.
+    process.env.FASTBOOT_NEW_BUILD = true;
```
kratiahuja: I wanted this variable to be set early on. Isn't `included` called after `init`?

index.js
```diff
@@ -79,67 +87,133 @@ module.exports = {
...
-    var fastbootBuild = new FastBootBuild({
-      ui: this.ui,
+  _buildFastbootConfigTree : function(tree) {
+    var env = this.app.env;
+    var config = this.project.config(env);
+    var fastbootConfig = config.fastboot;
+    // do not boot the app automatically in fastboot. The instance is booted and
+    // lives for the lifetime of the request.
+    if (config.hasOwnProperty('APP')) {
+      config['APP']['autoboot'] = false;
```
kratiahuja: Just prefer string keys :)

lib/build-utilities/migrate-initializers.js
```diff
@@ -0,0 +1,127 @@
...
+
+    const fastbootDirPath = path.join(rootPath, FASTBOOT_DIR);
+    // check if fastboot/app exists
+    if(!existsSync(fastbootDirPath)) {
+      fs.mkdirsSync(fastbootDirPath);
+    }
+
+    // copy over app/initializers/fastboot and app/instance/initializers/fastboot
+    fastbootInitializerTypes.forEach((fastbootInitializerType) => {
+      var srcFastbootPath = path.join(rootPath, 'app', fastbootInitializerType, 'fastboot');
```
kratiahuja: Yup definitely yes :)

package.json
```diff
@@ -30,9 +31,9 @@
     "ember-cli-babel": "^5.1.7",
     "ember-cli-eslint": "^3.0.2",
     "ember-cli-version-checker": "^1.3.1",
+    "exists-sync": "0.0.4",
```
kratiahuja: I thought `fs.existsSync` is deprecated. The readme of exists-sync says this: https://github.com/ember-cli/exists-sync#exists-sync

rwjblue: `fs.existsSync` was deprecated along with `fs.exists` a long way back in https://github.com/nodejs/node/commit/7b5ffa46fe4d2868c1662694da06eb55ec744bde, but then `fs.existsSync` was undeprecated in https://github.com/nodejs/node/commit/7b5ffa46fe4d2868c1662694da06eb55ec744bde.

kratiahuja: Aah didn't know that. Will use `fs.existsSync` and update readme for `exists-sync` too.

mazondo:
Anything I can help with to push this over the edge - we're keen on using it in our fastboot migration

stefanpenner:
> Anything I can help with to push this over the edge - we're keen on using it in our fastboot migration

We have a meeting tomorrow afternoon/evening to get this shipped.

scottmessinger:
I tried out this branch and it's working great! The one problem I've encountered is handling `app.import`. The discussion in this issue seems to focus on addons. In my app, addons aren't really the problem -- vendor libraries are. I'm importing libraries like interact.js, CKEditor, iscroll, etc and I had to go into each file in `/vendor` and add `if (window.document !== undefined) {...}`. Upgrading the libraries in the future will be harder since I'll have to copy and paste the library into the `if` block.

In addition to converting my vendor libs over, I tried getting `ember-fetch` working with this branch using the solution proposed here `https://gist.github.com/kratiahuja/d22de0fb1660cf0ef58f07a6bcbf1a1c` and working from the PR here: https://github.com/tomdale/ember-network/pull/20/files. I found the instructions in the gist to be somewhat confusing. I really appreciate @kratiahuja for drafting that up and I acknowledge that some of the complexity might be necessary. I'm hoping it can avoided and tucked behind a more declarative API.

Before this PR, the `if (isFastBoot){...} else {....}` was clear because it was a simple branch -- do the opposite if Fastboot is there. Now, there's a wildly different way to include a lib for Fastboot vs not including it. For instance, when including for the browser, a file can be included without being evaled if it's filtered out in `treeForVendor`. A couple steps, but they're logical. However, the opposite -- only include a file for fastboot is a completely different technique, requiring knowledge of the fastboot manifest.

In working with `ember-fetch` and my own app, what I wanted was a much simpler, more declarative way of filtering libs for Fastboot: `app.import('libname.js', {loadInFastBoot: false})`. I'm not sure if an API like that is possible, but I think it would be a helpful improvement.

kratiahuja:
Thanks @scottmessinger for your feedback. I appreciate it. 

> I had to go into each file in /vendor and add if (window.document !== undefined) {...}. Upgrading the libraries in the future will be harder since I'll have to copy and paste the library into the if block.

You could do this using `broccoli-stew` as well to avoid copy and paste issues in future. You could use an in-repo addon too for such a purpose.

> In working with ember-fetch and my own app, what I wanted was a much simpler, more declarative way of filtering libs for Fastboot: app.import('libname.js', {loadInFastBoot: false}). I'm not sure if an API like that is possible, but I think it would be a helpful improvement.

I agree with your thoughts on this. Did you read the build spec [here](https://gist.github.com/kratiahuja/fd073007e10abb9db0a2ec42bc1d7c17) ? There is an unresolved item about `app.import` that we can address in future. The goal of this PR is to come up with a minimal set of changes required to make the build more performant and make it to cut FastBoot major release version. 

> Before this PR, the if (isFastBoot){...} else {....} was clear because it was a simple branch -- do the opposite if Fastboot is there. Now, there's a wildly different way to include a lib for Fastboot vs not including it. For instance, when including for the browser, a file can be included without being evaled if it's filtered out in treeForVendor. A couple steps, but they're logical. However, the opposite -- only include a file for fastboot is a completely different technique, requiring knowledge of the fastboot manifest.

I understand that but the new build schema is different than the current one. The idea is instead of forking assets for different environments you want to load the browser assets and additional fastboot assets in fastboot environment. So you will need to load the fastboot assets in a different way. An app that is being server side rendered will nneed to have knowledge of fastboot manifest. I don't think that is wrong assumption. 

> I found the instructions in the gist to be somewhat confusing. I really appreciate @kratiahuja for drafting that up and I acknowledge that some of the complexity might be necessary. I'm hoping it can avoided and tucked behind a more declarative API.

The gist is very much in early draft stages. It was meant to iron out the different usecases where we need a migration path forward. As mentioned in https://github.com/ember-fastboot/ember-cli-fastboot/issues/360 there needs to be a migration guide to help addon owners.

scottmessinger:
@kratiahuja Thanks for the response! I'm excited you're open to letting `app.import` filter fastboot assets.

> An app that is being server side rendered will need to have knowledge of fastboot manifest. I don't think that is wrong assumption.

While I do think it's completely reasonable an app/addon to have knowledge of **fastboot**, I don't follow why an app/addon needs an understanding of the **fastboot manifest**. When "you want to load the browser assets and additional fastboot assets in fastboot environment", I think you'd only need a higher level api (like `app.import('fileName', {loadInFastboot: false})`). While having lower level primitives (involving the fastboot manifest) is a sound idea, in my mind, it's better to keep those primitives private in order to reduce the surface area of the public API one needs to learn just to get their app/addon working in fastboot. That all said, you know way, way more about fastboot than I, so maybe requiring addon/app developers to learn the fastboot manfest and more Broccoli callbacks is important and necessary.

Totally understand about the gist being an early draft -- I figured as much and really appreciate that you wrote it up.

stefanpenner:
> app.import('fileName', {loadInFastboot: false})).

I believe this already exists, or at-least the machinery to make it work.  

```js
app.import(thing, { outputFile: 'fastboot/some-file-only-required-in-input.js' }
```

But this discussion should likely find a new issue / rfc or something.

stefanpenner:
Awesome job @kratiahuja!

stefanpenner:
We should cut a beta today, then ask for community help in upgrading existing add-ons.

Burn down issue for addons that need <3 https://github.com/ember-fastboot/ember-cli-fastboot/issues/387

Issue #358, opened 2017-02-25
disable fastboot with unified serving


mfazekas:
#356 unifies fastboot serving. Apps having `ember-cli-fastboot` in their `package.json` will get fastboot render with ember 2.12 or later when pointing browser to `http://locahost:4200`.

It would be great to provide some options to get non fastboot render. 

Usecases:
- i see some issue in fastboot version and want to check if i see the same on regular version
- i've started fastboot adoption, but it's work in progress so i still want to default to non fastboot version for any feature work
- browser vs node.js dev experience - debugging

Current options, and it's limitations:
- remove `ember-cli-fastboot` from package.json. 
    Issues:
    - requires restart
    - if code uses `service:fastboot` we get `Attempting to inject an unknown injection: 'service:fastboot'`
- set `FASTBOOT_DISABLED`=TRUE
    Issues:
    - requires restart
    - currently it's half baked as does disable package.json generation, but doesn't disable fastboot serving, which results in `Couldn't find package.json error from EmberApp.readPackageJSON`

See:
https://github.com/ember-fastboot/ember-cli-fastboot/pull/356#issuecomment-281784396

mfazekas:
I think the best would be if apps could have to option to default to non fastboot version in general, and we would have the option to get the fastboot version via query parameter for example

stefanpenner:
What if there was a `?_fastboot=true|false` optional query param. That way you could toggle on a per request basis (to aide in quick debugging)

kratiahuja:
In the fastboot meeting yesterday we decided to explore either the query param option as @stefanpenner suggested or have an process environment variable. I will address this once that PR is merged.

fotinakis:
@kratiahuja +1 for environment variable, especially so there's no chance of a malicious user disabling fastboot in a prod app.

rwjblue:
FWIW, what we do here is generally unrelated to how it will work in production. The `ember serve` server is not intended for production use (you'd generally use `fastboot-app-server` or similar system that leverages the `fastboot-express-middleware` independently).

fotinakis:
Great, that's good to know, I haven't actually run fastboot in production yet.

stefanpenner:
> @stefanpenner suggested or have an process environment variable.

@kratiahuja seems like we should do both.

kratiahuja:
@stefanpenner +1 to both

josemarluedke:
[...]

PR #371, opened 2017-04-08
Enable a way to disable fastboot serving

kratiahuja:
**TL;DR: Fixes [issue#358](https://github.com/ember-fastboot/ember-cli-fastboot/issues/358)**

This PR provides a way to disable fastboot serving with `ember serve` via the following:
- [x] Runtime query parameter `fastboot` when using `ember serve`. In order to disable fastboot serving on per request basis use: `http://localhost:4200/?fastboot=false`.
- [x] Provides a way to disable fastboot serving via environment variable `FASTBOOT_DISABLED`. The same environment variable is used for disabling double builds.
- [x] Updates README.

cc: @rwjblue @stefanpenner @mfazekas @danmcclain

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,13 @@ ember install ember-cli-fastboot
 * `ember fastboot --serve-assets`
 * Visit your app at `http://localhost:3000`.
 
-**Note**: If your app is running ember-cli v2.12.0-beta.1+, you can just use `ember serve` instead of `ember fastboot --serve-assets`.
+### With `ember-cli` version 2.12.0-beta.1 and above
+If your app is running ember-cli v2.12.0-beta.1+, you can just use `ember serve` instead of `ember fastboot --serve-assets` and visit at `http://localhost:4200/`.
+
+Optionally you can even disable the fastboot serving at runtime using the `fastboot` query parameter. Example to turn off fastboot serving,
+visit your app at `http://localhost:4200/?fastboot=false`. If you want to turn on fastboot serving again, simply visit at `http://localhost:4200/?fastboot=true` or `http://localhost:4200/`.
+
+You can even disable serving fastboot with `ember serve` using an environment flag: `FASTBOOT_DISABLED=true ember serve`. If you have disabled building fastboot assets using the same flag as described [here](https://github.com/ember-fastboot/ember-cli-fastboot#double-build-times-and-no-incremental-builds), remember to also disable serving fastboot assets when using `ember serve`.
 
 You may be shocked to learn that minified code runs faster in Node than
 non-minified code, so you will probably want to run the production
@@ -392,7 +398,7 @@ inside of the sandbox environment. For this reason, it's encouraged to [disable 
 Due to limitations in Ember CLI, builds take twice as long to generate the
 second set of FastBoot assets. This also means incremental builds with
 live reload don't work either. This aims to be resolved by FastBoot 1.0.
-In the mean time, we introduce a short-circuit evironment flag to not do
+In the mean time, we introduce a short-circuit environment flag to not do
 a FastBoot build:
 
 ```
--- a/index.js
+++ b/index.js
@@ -152,12 +152,13 @@ module.exports = {
       // that version contains API to hook fastboot into ember-cli
 
       app.use((req, resp, next) => {
+        var fastbootQueryParam = (req.query.hasOwnProperty('fastboot') && req.query.fastboot === 'false') ? false : true;
+        var enableFastBootServe = !process.env.FASTBOOT_DISABLED && fastbootQueryParam;
         var broccoliHeader = req.headers['x-broccoli'];
         var outputPath = broccoliHeader['outputPath'];
 
-        if (broccoliHeader['url'] === req.serveUrl) {
+        if (broccoliHeader['url'] === req.serveUrl && enableFastBootServe) {
           // if it is a base page request, then have fastboot serve the base page
-          // TODO(future): provide a way to turn this off without needing to uninstall this addon
           if (!this.fastboot) {
             // TODO(future): make this configurable for allowing apps to pass sandboxGlobals
             // and custom sandbox class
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -33,7 +33,13 @@ ember install ember-cli-fastboot
 * `ember fastboot --serve-assets`
 * Visit your app at `http://localhost:3000`.
 
-**Note**: If your app is running ember-cli v2.12.0-beta.1+, you can just use `ember serve` instead of `ember fastboot --serve-assets`.
+### With `ember-cli` version 2.12.0-beta.1 and above
+If your app is running ember-cli v2.12.0-beta.1+, you can just use `ember serve` instead of `ember fastboot --serve-assets` and visit at `http://localhost:4200/`.
+
+Optionally you can even disable the fastboot serving at runtime using the `fastboot` query parameter. Example to turn off fastboot serving,
+visit your app at `http://localhost:4200/?fastboot=false`. If you want to turn on fastboot serving again, simply visit at `http://localhost:4200/?fastboot=true` or `http://localhost:4200/`.
```
bcardarella: would it make more sense to use `-fastboot` instead or would we consider `fastboot` namespaced enough?

kratiahuja: I think it is going to be more error prone to have `-` in the query parameters. Developers can accidentally forget about it.

bcardarella: The point would be to avoid potential collisions and not limit the QP options for the consuming application. That being said, I can't think of a reason someone would want to create a `fastboot` QP

danmcclain: Instead of it being a switch, I would think it would be a per-request option. if you remove the `fastboot=false` QP, you'd get a FastBoot response again.

kratiahuja:
I guess we can assume it now becomes a reserved keyword if apps use fastboot :)

README.md
kratiahuja: @bcardarella I guess we can assume `fastboot` will be a reserved QP with this PR :)

index.js
```diff
@@ -152,10 +152,12 @@ module.exports = {
...
       app.use((req, resp, next) => {
+        var fastbootQueryParam = (req.query.hasOwnProperty('fastboot') && req.query.fastboot === 'false') ? false : true;
+        var enableFastBootServe = !process.env.FASTBOOT_DISABLED && fastbootQueryParam;
         var broccoliHeader = req.headers['x-broccoli'];
         var outputPath = broccoliHeader['outputPath'];
 
-        if (broccoliHeader['url'] === req.serveUrl) {
+        if (broccoliHeader['url'] === req.serveUrl && enableFastBootServe) {
           // if it is a base page request, then have fastboot serve the base page
           // TODO(future): provide a way to turn this off without needing to uninstall this addon
```
mfazekas: i think this TODO comment is addressed now

index.js
```diff
@@ -152,10 +152,12 @@ module.exports = {
       // that version contains API to hook fastboot into ember-cli
 
       app.use((req, resp, next) => {
+        var fastbootQueryParam = (req.query.hasOwnProperty('fastboot') && req.query.fastboot === 'false') ? false : true;
+        var enableFastBootServe = !process.env.FASTBOOT_DISABLED && fastbootQueryParam;
```
mfazekas: I think it would be nice if we'd start with `FASTBOOT_DISABLED=true` we'd still served can try fastboot for `http://localhost:4200/?fastboot=true` the usecase is non fastboot first development but we want to try stuff in fastboot too.

mfazekas: There is also comment on line 119 that needs to be changed i think
```
FASTBOOT_DISABLED is a pre 1.0 power user flag to...
```

mfazekas:
Looks a works great to me. Ideally there would be a way so apps can specify (ideally in their config) so that they prefer non fastboot version, and they only want fastboot enabled for `/?fastboot=true`. The usecase is that while we use fastboot, we prefer doing development on non fastboot version as debugging etc is much easier. So we want to default to non fastboot but want fastboot on an adhoc manner. Kind of reverse the current default. 
I think double builds are less of a concern for such usecase.

index.js
kratiahuja: @mfazekas If you set `FASTBOOT_DISABLED`, irrespective of the query param you will never be served via fastboot. The environment variable is temporary and will go away after the double builds are resolved.

kratiahuja:
@mfazekas 
> So we want to default to non fastboot but want fastboot on an adhoc manner. Kind of reverse the current default.

I disagree that the default behavior should be non fastboot version if you want to serve your app in production using fastboot. One of the motivations of unifying `ember fastboot --serve-assets` with `ember serve` is so that apps can debug fastboot issues sooner rather than later. I **strongly** believe that the default behavior should be apps with `ember-cli-fastboot` should be served via fastboot unless they disable fastboot serving via environment variable or on per request basis.

mfazekas:
@kratiahuja sorry i might not have been clear. I agree that default should be fastboot on and option to turn it off which is the PR implements. 

For some projects it would be great to be able to have a reverse mode, where fastboot is off by default but can enabled with env/on request basis. For example if the projects fastboot support is not 100% complete, such settings would allow to integrate ember-cli-fastboot and some fastboot related changes to project without affecting other developers, until the project's fastboot support is complete.

But there is also question of dev experience: bugs in js code will be printed in nodejs env instead of chrome debugger. For some projects developer productivity might have higher priority than surfacing potential fastboot issues sooner.

kratiahuja:
Doesn't the `FASTBOOT_DISABLED` environment flag solve that for you? We can chose to keep it even outside the scope of the double build or introduct ember-cli config or something. Though I don't think it is tied to this PR.

kratiahuja:
@rwjblue can you review this PR please? This is blocking the release of `ember-cli-fastboot` with unified serve.

mfazekas:
@kratiahuja sure it's not neccesary to handle it here i just wanted to mention, i'll open another issue for that.

kratiahuja:
Thank you @rwjblue!

PR #373, opened 2017-04-11
Updated information around node-inspector

ryanone:
[The fix](https://github.com/node-inspector/node-inspector/commit/2e5309f75099753740c4567e17fd79ee27885d71) for the [node-inspector issue](https://github.com/node-inspector/node-inspector/issues/905), can now be leveraged in versions ~1.0.0.  I've tried this out and it works fine.  I've updated the README to reflect this.


rwjblue:
Thank you!

PR #374, opened 2017-04-15
Expose API for fastboot addons to update fastboot manifest.

kratiahuja:
## Motivation
Currently, FastBoot runs two ember builds to fork and build fastboot assets. When it runs the fastboot build it sets an environment variable `process.env.EMBER_CLI_FASTBOOT` to be `true`.  At present a lot of addons are relying on this environment variable to append libraries into the fastboot assets (example `ember-network`).

Moreover another usecase is to allow addons like `ember-engines` to write and load additional assets in the FastBoot sandbox.

With #360, this environment variable will no longer be exposed or valid. Therefore, we want to expose an API from `ember-cli-fastboot` to allow other fastboot compatible addons to be able to provide paths under `dist` from where additional `vendor` or `app` specific assets should be loaded in sandbox. With addons being able to leverage this API when the build issue is resolved they will automatically become backward compatible.

## How does it work
In order for an addon to load additional assets in the sandbox (in any order), they need to either update `vendorFiles` or `appFiles` key in the fastboot manifest. Therefore, they will define `updateFastBootManifest` hook in their `index.js` which takes in the `manifest` as input, updates the manifest and returns it back:
```javascript
   updateFastBootManifest: function(manifest) {
      manifest.vendorFiles.unshift(<add some path>);
      manifest.appFiles.push(<add some path>);
     
     return manifest;
  }
```

An addon can decide to update the manifest enteries however they like. If the `manifest` key is not returned, the build will fail with an error message as : `<addon name> did not return the updated manifest from updateFastBootManifest hook.`

**Example**: See [`ember-fastboot-build-example`](https://github.com/kratiahuja/ember-fastboot-build-example/blob/master/index.js#L11) addon for a simple example of how to load assets

## TODO
- [ ] Update README with instructions and example of how to use the new API
- [ ] Update fastboot website with new API

cc: @rwjblue @tomdale @danmcclain @ryanone @tsubomii @arjansingh @simonihmig

### DIFF
--- a/index.js
+++ b/index.js
@@ -177,7 +177,7 @@ module.exports = {
           // forward the request to the next middleware (example other assets, proxy etc)
           next();
         }
-      })
+      });
     }
   },
 
--- a/lib/broccoli/fastboot-config.js
+++ b/lib/broccoli/fastboot-config.js
@@ -115,6 +115,20 @@ FastBootConfig.prototype.readAssetManifest = function() {
   }
 };
 
+FastBootConfig.prototype.updateFastBootManifest = function(manifest) {
+  this.project.addons.forEach(function(addon) {
+    if (addon.updateFastBootManifest) {
+      manifest = addon.updateFastBootManifest(manifest);
+
+      if (!manifest) {
+        throw new Error(`${addon.name} did not return the updated manifest from updateFastBootManifest hook.`);
+      }
+    }
+  });
+
+  return manifest;
+}
+
 FastBootConfig.prototype.buildManifest = function() {
   var appFileName = path.basename(this.outputPaths.app.js).split('.')[0];
   var appFile = 'fastboot/' + appFileName + '.js';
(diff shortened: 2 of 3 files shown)


kratiahuja:
Travis is failing due to unrelated [ember-data issue](https://github.com/emberjs/data/issues/4927)

kratiahuja:
@rwjblue and I discussed over slack. We should also handle usecase where you may want to prepend to `vendorFiles` or `appFiles`. 

```js
updateFastBootManifest(manifest) {..}
```
you get manifest as input (which maybe an updated version from prev addon) and you update and return and at the end we write it back

Going to update this PR.

kratiahuja:
@rwjblue @simonihmig updated PR and description with the discussed changes. Also tested it with multiple addons to make sure ordering is maintained as well.

kratiahuja:
@rwjblue thank you! :)

PR #391, opened 2017-05-17
Use relative protocol for redirects

josemarluedke:
This allow apps that use https in the load balancer to redirect correctly. Currently if you have a https in the load balancer, the protocol that the express server will see is http, so when redirecting it will also use http instead of https. Now with relative protocol, it will use https instead.

I had a conversation in slack about it, and I know other people had issues like this.

### DIFF
--- a/app/locations/none.js
+++ b/app/locations/none.js
@@ -38,9 +38,8 @@ export default Ember.NoneLocation.extend({
         let isTransitioning = currentPath !== path;
 
         if (isTransitioning) {
-          let protocol = get(this, '_request.protocol');
           let host = get(this, '_request.host');
-          let redirectURL = `${protocol}://${host}${path}`;
+          let redirectURL = `//${host}${path}`;
 
           response.statusCode = this.get('_redirectCode');
           response.headers.set('location', redirectURL);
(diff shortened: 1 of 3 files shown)


kratiahuja:
@josemarluedke Is it possible to add tests for this?

josemarluedke:
@kratiahuja we had tests for this, but I did a rebase a while ago and these tests where gone.

josemarluedke:
The tests were removed here: https://github.com/ember-fastboot/ember-cli-fastboot/pull/398 

I will try to get them back.

kellyselden:
yikes it looks like lots of legit tests were removed on accident.

kratiahuja:
I think I meant to delete all tests using `ember fastboot --serve-assets` . However, I didn't realize these tests weren't migrated to use `ember serve`.

josemarluedke:
@kratiahuja I added `fastboot-location-config-test` back. I will get another PR with failing test for the other deleted file.

kratiahuja:
@kellyselden can you please take a look at this PR? I believe you know location/none stuff well.

kellyselden:
LGTM

Issue #394, opened 2017-05-18
Clean up and delete `ember fastboot` command


kratiahuja:
Once we release `ember-cli-fastboot@1.0.0`, after a few releases we should drop support for `ember fastboot` command and delete the command. 

Following are the things that need to be removed:
- [ ] Remove the [`ember fastboot`](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/lib/commands/fastboot.js) command
- [ ] Remove the TODOs from `index.js` [here](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L36) and other places in `index.js`
- [ ] Update all the tests to not test against `ember fastboot` command (ie test description stating "with fastboot command"). Example test is [here](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/test/error-handler-test.js#L12)

This should good for new contributors as well.

PR #398, opened 2017-05-19
Kill ember fastboot command

kratiahuja:
Fixes #https://github.com/ember-fastboot/ember-cli-fastboot/issues/394 

cc: @stefanpenner


kratiahuja:
we can fix appveyor as part of #392 PR.

PR #410, opened 2017-06-01
[BUGFIX] fix build stability

stefanpenner:
This prevents fastboot from constantly writing its package.json if it has not changed.

- [x]  tests

### DIFF
--- a/lib/broccoli/fastboot-config.js
+++ b/lib/broccoli/fastboot-config.js
@@ -7,18 +7,21 @@ const md5Hex = require('md5-hex');
 const path   = require('path');
 const Plugin = require('broccoli-plugin');
 
+const stringify = require('json-stable-stringify');
+
 const LATEST_SCHEMA_VERSION = 2;
 
 module.exports = class FastBootConfig extends Plugin {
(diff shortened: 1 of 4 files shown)


kratiahuja:
LGTM. I think there are existing tests around the package.json. What other tests did you have in mind?

stefanpenner:
> I think there are existing tests around the package.json

Unit tests, the existing ones are too high level to add proper coverage (and also quite slow and brittle)

stefanpenner:
@kratiahuja I have updated them. I would likely recommend we transition more and more of the tests to more targeted tests like this. Rather then the end to end tests we have now. I feel we have better confidence with the tests, easier to pin-point issues, easier to add additional scenarios etc, then the existing approach.

simonihmig:
Tried this also, and can confirm this works perfectly.

For the record: this also fixes the issue that CSS hot reloading was not working correctly, the browser would always do a full reload even if just some CSS/Sass files changed.

stefanpenner:
I'll release rc2 when I get to work

stefanpenner:
released as v1.0.0-rc.2 🎉

PR #411, opened 2017-06-01
[BUGFIX] Fix reading config for browser environment

### LINKED ISSUE #409: Issues with storeConfigInMeta being false
It seems like there is a new error for apps using `storeConfigInMeta = false`. I think it has to do with explicitly checking for the meta-tag prior to the code that generates this error: `Uncaught Error: Could not read config from meta tag with name "<app-name>/config/environment".`

kratiahuja:
Fixes #409 

Push whatever ember-cli gives instead of re-creating the same.

cc: @stefanpenner

### DIFF
--- a/index.js
+++ b/index.js
@@ -77,15 +77,14 @@ module.exports = {
     }
 
     // if the fastboot addon is installed, we overwrite the config-module so that the config can be read
-    // from meta tag for browser build and from Fastboot config for fastboot target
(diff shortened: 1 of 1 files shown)


kratiahuja:
Reason for not setting `storeConfigInMeta` to be false explicitly (which would not require us to have `config-module` override) in this addon via `app.options` is because if there is an addon that ends up setting it to true and runs after `ember-cli-fastboot`, it will break fastboot.

PR #412, opened 2017-06-01
checking undefined directly

bekzod:


stefanpenner:
I believe the reason this is typeof is that Fastboot may not be defined

PR #415, opened 2017-06-02
removed deprecated things

bekzod:
noticed that deprecated `until: '0.9.9'`  haven't been removed when `v1.0.0-rc.2` released, so this removes them

### DIFF
--- a/app/services/fastboot.js
+++ b/app/services/fastboot.js
@@ -61,8 +61,6 @@ const Shoebox = Ember.Object.extend({
 });
 
 const FastBootService = Ember.Service.extend({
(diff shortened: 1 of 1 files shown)


stefanpenner:
@kratiahuja do we want to leave this around awhile longer to provide helpful errors or..

kratiahuja:
Yes, I would prefer we remove post 1.0.0. We can update the `until` property to be 1.0.0.

stefanpenner:
> We can update the until property to be 1.0.0.

Until likely doesn't need to change, it merely states a lower bound for actual removal.

bekzod:
ok closing this then :P

PR #416, opened 2017-06-02
refactored shoebox object

bekzod:
this is more of suggestion and does not solve any issues probably :P
* avoids using ember get/set
* doesn't set cached values to itself

### DIFF
--- a/app/services/fastboot.js
+++ b/app/services/fastboot.js
@@ -27,26 +27,33 @@ const RequestObject = Ember.Object.extend({
 });
 
 const Shoebox = Ember.Object.extend({
+
+  init() {
+    this._super(...arguments);
+
+    this.isFastBoot = this.get('fastboot.isFastBoot');
+    this.fastbootInfo = this.get('fastboot._fastbootInfo');
+    this._cache = Object.create(null);
+  },
+
   put(key, value) {
-    assert('shoebox.put is only invoked from the FastBoot rendered application', this.get('fastboot.isFastBoot'));
+    assert('shoebox.put is only invoked from the FastBoot rendered application', this.isFastBoot);
     assert('the provided key is a string', typeof key === 'string');
 
-    let fastbootInfo = this.get('fastboot._fastbootInfo');
+    let fastbootInfo = this.fastbootInfo;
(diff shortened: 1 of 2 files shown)


kellyselden:
I'm having trouble seeing the point of this. What's wrong with using get/set? If it avoids putting more state on the shoebox for people to abuse, that seems like a good thing?

bekzod:
ember get/set a bit slower no ? :P, yeah but main reason was to store cached values in separate obj which makes things a bit cleaner I guess :)

kratiahuja:
The shoebox is itself a cache today. Why create it as a property?

bekzod:
@kratiahuja can you look at removed test, is that desired behavior ? I would expect not :) like tampering with cached values etc., am I wrong ?

kratiahuja:
@bekzod I don't think we should remove that test. Why are you removing that test?

kellyselden:
The main advantage to using a new object is you free up using the keys `retrieve` and `put`.

bekzod:
also maybe `retrieve and put` could be renamed just get and set :P ?

kratiahuja:
That's going to be a breaking change for apps that are using this API.

I am still not understanding the motivation behind this change.

bekzod:
I didn't mean now, I meant `later`, motivation was to just introduce cleaner way storing cached values storing cached values to itself looks a but sloppy, 

* this frees up `retrieve` and `put` so avoids overwriting them
* avoids tampering with cached values  `shoebox.set('foo', 'bar');` doesn't change cache

if you thing those are not valid points, free to close

kratiahuja:
`put` is only supposed to be running in when your app runs in the fastboot environment. `retrieve` is to be used for app running in the browser to re-hydrate on the same api responses that the same app on the server side chose to put/use. 

An app can chose to override the values at any point during its lifecycle run in the fastboot environment. Once the shoebox is locked in and send to the browser, there ideally don't need to override since they only would want to read. If they need to override, for any logical reason (example: use the shoebox value but in background re-fetch the data and store it) I don't think that is wrong. It's really up to the app how they would like to use it and I am not seeing a strong reason for us to not allow them to do that.

PR #417, opened 2017-06-03
create shoebox object with empty proto

bekzod:


kellyselden:
This needs to consider that it will no longer be compatible with fastboot <= 1.0.0-rc.6 because it will be missing the compat fix from https://github.com/ember-fastboot/fastboot/pull/144.

PR #418, opened 2017-06-04
remove global flag.

stefanpenner:
Addons can check via:

this.project.addons.find(addon => addon.name === 'ember-cli-fastboot'):
this.project.findAddonByName('ember-cli-fastboot');

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,10 +51,6 @@ module.exports = {
     this._appRegistry = app.registry;
     this._name = app.name;
 
-    // set a environment variable to allow addons to use `fastboot-filter-initializers`
-    // for old versions.
-    process.env.FASTBOOT_NEW_BUILD = true;
-
     migrateInitializers(this.project);
   },
 


kratiahuja:
This was introduced for a corner usecase for `ember-cli-head` to be backward compatible: https://github.com/ronco/ember-cli-head/blob/master/index.js#L13

kratiahuja:
Though now that I think about it, this flag would not be needed if `treeForApp` in `ember-cli-head` is refactored.

stefanpenner:
https://github.com/ronco/ember-cli-head/pull/30

stefanpenner:
@kratiahuja ah, that would be even better. But if we need to check explicit things, (like capability) a version check may be more appropriate then a global flag.

But if it can be refactored away entirely, we should take that approach.

kratiahuja:
Yeah I don't think we need the `if..else` really...

Issue #421, opened 2017-06-08
treeForFastBoot does not work for nested addons


simonihmig:
Not sure if this should be considered a bug or is so by design: the Fastboot tree is build by iterating over all addons from `this.project.addons` ([source](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L115)), which only includes the project's own addons, not any nested ones. 

Here is a real world example where this approach fails:
* ember-cli-head (as of its latest release 0.3.0) has an instance-initializer in its `/fastboot` folder (https://github.com/ronco/ember-cli-head/blob/master/fastboot/instance-initializers/head.js)
* ember-cli-meta-tags (4.0.0) uses and has ember-cli-head as a dependency
* when only ember-cli-meta-tags is added to a project, ember-cli-head is not part of `this.project.addons`, so its initializer is **not** added to `appName-fastboot.js`
* the whole thing does not work anymore in FastBoot (doesn't render meta tags)

Not sure how to solve this...
* recursively fetch all nested addons and build their fastboot tree as well?
* change ember-cli-meta-tags to use `addAddonToProject` in its default blueprint to add ember-cli-head as a direct dependency?

kratiahuja:
I wouldn't consider this as a bug since we are relying on the addons `this.project.addons` to tell us the addons in the eco-system. `ember-cli` builds `this.project.addons`. I wonder if `peerDependencies` would solve it? How was this working in pre-Fastboot 1.0 world?

> recursively fetch all nested addons and build their fastboot tree as well?
You could have n level deep addons. I am not sold on this idea yet.

simonihmig:
> ember-cli builds this.project.addons

I don't think this is true. Nested addons have their caveats for sure, but in general work. All the browser code of nested addons is correctly added to the build, in this case for example the browser initializers of ember-cli-head!

As @rwjblue just told me on Slack regarding this, each addon has an own `.addons` array, so you could iterate over those nested addons as well. And I believe in some form ember-cli must do so!?

kratiahuja:
My concern is recrusively looping through is going to cause perf issues.

kratiahuja:
Just brainstorming:

Why not expose an API where the dependent addon registers its dependent addons to `ember-cli-fastboot`? This way it is the responsibility of the `ember-cli-meta-tags` to moves its dependencies as into a flatter structure. This has a caveat though: the flatter structure could have repeated addons and I think the last one will win.

[...]

PR #422, opened 2017-06-08
Allow in-repo addons (or scoped addons) to have `fastboot` trees.

rwjblue:
Prior to this change only "normal" package based addons could have a `fastboot` folder auto discovered.

/cc @kratiahuja

### DIFF
--- a/index.js
+++ b/index.js
@@ -109,12 +109,11 @@ module.exports = {
    */
   _getFastbootTree: function() {
     const appName = this._name;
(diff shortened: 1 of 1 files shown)


simonihmig:
Nice! FWIW when debugging #421 I saw that ember-cli-mocha resulted in a wrong path, because of its slightly unusual addon name (which was fixed recently, see https://github.com/ember-cli/ember-cli-mocha/commit/5827cb43c23592bb757f13b42783d4e6b72f385d). This change should fix that as well...

Issue #420, opened 2017-06-08
Semantics of treeForFastBoot


simonihmig:
Just stumbled upon this when looking at the code: I am not sure if the current semantics of the `treeForFastBoot` hook are what users would expect, or makes most sense.

Currently any addon's `/fastboot` tree is taken ([source](https://github.com/ember-fastboot/ember-cli-fastboot/blob/6bc10810c40cb0d08aa9078d4dc7d9f6953984b2/index.js#L120-L121)) and the (optional) output of calling `treeForFastBoot` on the addon is merged *on top* of that ([source](https://github.com/ember-fastboot/ember-cli-fastboot/blob/6bc10810c40cb0d08aa9078d4dc7d9f6953984b2/index.js#L128)). Which is ok when you want to modify/extend the tree. But using a Funnel in that hook seems to be pointless.

My understanding of the other ember-cli's `treeFor` hooks like `treeForPublic` is that only the tree that those hooks return is taken into account (so Funnel works). And I feel the same semantics should probably be also applied to the `treeForFastBoot` hook, shouldn't they?

Happy to add a PR if we can agree on this!

kratiahuja:
What happens when you use a funnel? This hook simply returns a `broccoli` tree so I imagine funnel should also work.

simonihmig:
@kratiahuja actually I have not tested this, but by looking at the code I would say that because the whole ("unfunneled") tree is added regardless of the output of the hook (see https://github.com/ember-fastboot/ember-cli-fastboot/blob/6bc10810c40cb0d08aa9078d4dc7d9f6953984b2/index.js#L120-L121), you would still end up with having the whole tree, even when the hook returns a reduced (funneled) one!? 

Since the whole, unfunneled tree is part of the `mergeTrees` call, you cannot remove anything from it. Or do I miss something?

simonihmig:
Btw, the output of the hook is assigned to `additionalFastBootTree`, so that says it all, doesn't it? :)

rwjblue:
I addressed this in #423

PR #423, opened 2017-06-08
Ensure `treeForFastBoot` is called for arbitrarily nested addons.

rwjblue:
Previously, we only called `treeForFastBoot` on the projects direct dependencies.

* updates to call `treeForFastBoot` at each layer (just like `treeFor` does)
* Ensures that we do not double process the raw addons `fastboot` folder **and**
  the result of `treeForFastBoot`.

Fixes https://github.com/ember-fastboot/ember-cli-fastboot/issues/420
Fixes https://github.com/ember-fastboot/ember-cli-fastboot/issues/421

### DIFF
--- a/index.js
+++ b/index.js
@@ -98,7 +98,35 @@ module.exports = {
 
     return tree;
   },
+  
+  _processAddons(addons, fastbootTrees) {
+    addons.forEach((addon) => {
+      this._processAddon(addon, fastbootTrees);
+    });
+  },
+  
+  _processAddon(addon, fastbootTrees) {
+    // walk through each addon and grab its fastboot tree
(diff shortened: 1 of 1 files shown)


kratiahuja:
Thank you @rwjblue

simonihmig:
Just tested these changes with the addon combination mentioned in #421, works like a charm, and indeed fixes that problem! Thanks a lot!

rwjblue:
Thanks for reviewing @kratiahuja, and @simonihmig for confirming in the original failure scenario!

I won't have time to release tonight, not sure if @kratiahuja / @stefanpenner / or others do...

kratiahuja:
I'll release tonight!

kratiahuja:
ember-cli-fastboot@1.0.0-rc.3 is released!

PR #436, opened 2017-06-27
BUGFIX: Error logs only showing [Object object]

BlueRaja:
If an error-object that is not an `Error` is propagated, the FastBoot logs show

> There was an error running your app in fastboot. More info about the error:
> [Object object]

This fixes that issue.


kratiahuja:
LGTM. Can you add a test for this?

PR #437, opened 2017-06-30
Improve explanation for importing incompatible libraries

courthead:
A lot of the documentation here seems targeted at addon authors rather than regular users of Ember CLI. I don't think it was made clear to the latter group that they absolutely cannot use `ember-cli-build.js` anymore and instead must create an in-repo addon for the sole purpose of importing FastBoot-incompatible JS libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -356,13 +356,14 @@ export default Ember.Route.extend({
 
 ### Disabling incompatible dependencies
 
-There are two places where the inclusion of incompatible JavaScript libraries could
-occur:
+There are two places where the inclusion of incompatible JavaScript libraries could occur:
 
  1. `app.import` in the application's `ember-cli-build.js`
  2. `app.import` in an addon's `included` hook
+ 
+There is no way to exclude incompatible libraries imported from the application's `ember-cli-build.js`. Instead, you will need to create an in-repo addon and include your libraries from there.    
 
-You can include the incompatible Javascript libraries by wrapping them with a `FastBoot` variable check. In the browser, `FastBoot` global variable is not defined.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -356,13 +356,14 @@ export default Ember.Route.extend({
 
 ### Disabling incompatible dependencies
 
-There are two places where the inclusion of incompatible JavaScript libraries could
-occur:
+There are two places where the inclusion of incompatible JavaScript libraries could occur:
 
  1. `app.import` in the application's `ember-cli-build.js`
```
danmcclain: Seems like we should remove this line then?

kratiahuja:
@courthead Just an FYI, we are working on exposing APIs in Ember CLI to allow Ember apps to import FastBoot incompatible libraries. I am fine with merging this until that lands.

README.md
kratiahuja: +1 to @danmcclain comment.

tomdale:
I wonder if there's a way we could add an option to `app.import()` to disable it in FastBoot? This is a pretty annoying gotcha for app developers, as @courthead points out.

Maybe something like:

```js
app.import('node_modules/some-dep/dist/some-dep.js', { fastboot: false });
```

In this case, if the `fastboot` option is set to `false`, we can add a Broccoli transform that wraps the dependency in `if (typeof FastBoot === 'undefined') {` or whatever we expect addon authors to do manually. Or we could generate an empty stub module that overwrites the AMD module in the FastBoot build.

kratiahuja:
@tomdale Agree it is pretty annoying. We have an RFC in ember-cli that exposes a declarative API that will make it easier for app developers using app.import : https://github.com/ember-cli/rfcs/pull/108. Until then the workaround is to use in-repo addon

PR #443, opened 2017-07-14
Fix serving FastBoot when app has rootUrl defined

### LINKED ISSUE #441: App not being served by FastBoot when rootURL is set
I'm attempting to upgrade to FastBoot 1.0.0-rc.5 with Ember CLI version 2.12.2. The app's rootURL configuration is set to something other than the default, e.g. '/test/'. When I run the app using `ember serve`, it's not being served by FastBoot.

It looks like this is because `broccoliHeader['url']` does not match `req.serveUrl` in the check  [here](https://github.com/ember-fastboot/ember-cli-fastboot/blob/f0863abadd8226951dd6218b6c68cc44f6ee02a0/index.js#L241). `req.serveUrl` includes the rootURL but `broccoliHeader['url']` does not, e.g.
```
broccoliHeader['url']: /index.html
req.serveUrl: /test/index.html
```
and this causes the app to be served normally. (I also confirmed that the value of `enableFastBootServe` was true)

Additionally, I'm now unable to run the app using the `ember fastboot` command as I get an error message saying it's been removed.

kratiahuja:
Fixes #441 

We only need to check if `req.serveUrl` is defined to make sure it is a `index.html` request. `req.serveUrl` is set by the history middleware in `ember-cli` that runs before this middleware. For asset requests, `req.serveUrl` will be undefined so those requests will be served by `ember-cli`.

FastBoot reads `index.html` from dist folder so it doesn't care about the `rootUrl`.

cc: @stefanpenner @josemarluedke

### DIFF
--- a/index.js
+++ b/index.js
@@ -238,7 +238,7 @@ module.exports = {
         const broccoliHeader = req.headers['x-broccoli'];
         const outputPath = broccoliHeader['outputPath'];
 
-        if (broccoliHeader['url'] === req.serveUrl && enableFastBootServe) {
+        if (req.serveUrl && enableFastBootServe) {
           // if it is a base page request, then have fastboot serve the base page
           if (!this.fastboot) {
             // TODO(future): make this configurable for allowing apps to pass sandboxGlobals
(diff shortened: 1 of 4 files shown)


stefanpenner:
released as v1.0.0-rc.6 🎉

PR #448, opened 2017-07-19
Update fastboot, fastboot-express-middleware

kratiahuja:


stefanpenner:
```
 FastBootLocation
    ✓ should NOT redirect when no transition is called (719ms)
    ✓ should NOT redirect when intermediateTransitionTo is called (49ms)
    1) should redirect when transitionTo is called
    2) should redirect when replaceWith is called
```

app/locations/none.js
```diff
@@ -40,7 +40,7 @@ export default Ember.NoneLocation.extend({
         if (isTransitioning) {
           let protocol = get(this, '_request.protocol');
           let host = get(this, '_request.host');
-          let redirectURL = `${protocol}://${host}${path}`;
+          let redirectURL = `${protocol}//${host}${path}`;
```
stefanpenner: :+1:

PR #449, opened 2017-07-19
upgrade deps blocking new node versions

stefanpenner:
this will likely need some additional TLC... but leaving it here as WIP


kellyselden:
all deps but qunit were upgraded. qunit is the problem child right now.

PR #450, opened 2017-07-19
Add back "Use relative protocol for redirects"

josemarluedke:
cc/ @kellyselden.

Issue #379, opened 2017-04-24
transitionTo under https


zion03:
@zion03 commented on [Fri Apr 21 2017](https://github.com/ember-fastboot/fastboot/issues/135)

I'm trying to transition `fastboot` under the `https`, it's works with redirection to `http` version.

For example `application.js` route

```
export default Ember.Route.extend({
  model: function(param){
    this.transitionTo('login');
  }
});
```
request: `https://domain.com` -->  `http://domain.com:3001/login`
where `3001` fastboot port

How I can solve this problem?

danmcclain:
Do you have a proxy like nginx in front of FastBoot? I'm assuming so since you are using SSL. Make sure that you forward your headers, as FastBoot is receiving an http request from your proxy and doesn't know that it's actually an https request from the browser. I think you need to pass X-forwarded-proto like in this article https://www.digitalocean.com/community/tutorials/how-to-configure-nginx-with-ssl-as-a-reverse-proxy-for-jenkins

zion03:
More details about our production environment:

We are using Amazon ELB, and we use https listener on a custom port, f.e. 44443 and we terminate SSL on this loadbalancer. We send a request to https://example.com:44443, per Amazon documentation, ELB sets X-forwarded-proto to the same schema as original request has. https in this case case. ELB accesses fastboot via http

Actually we use nginx also, but it sits before ELB described above, and proxy_pass requests using https to this ELB.
We see the same redirect behavior in both cases even we request directly via ELB https://example.com:44443 or using via nginx proxy_pass

kratiahuja:
You should be able to use a middleware that redirects to https. See here: https://github.com/ember-fastboot/fastboot-app-server/issues/36

kratiahuja:
Going to close this as you can redirect via a middleware. Please feel free to reopen if this doesn't work for you.

devdemi:
@kratiahuja I have the same problem as @zion03 but middleware doesn't help me. 
My fastboot works behind nginx on port 3443 and when any `transitionTo` called fastboot will transfer to domain.com:3443 

Lines from my nginx config:
```
    location / {
        if ($http_x_forwarded_proto != 'https') {
            return 301 https://$host$request_uri;
        }
        #try_files @fastboot $uri $uri/ /index.html;
        resolver 8.8.8.8;
        proxy_pass https://$server_name:3443;
        proxy_redirect     off;

        proxy_set_header X-Forwarded-Host $host;
[...]

PR #454, opened 2017-07-21
Protocol double colon fix

krasnoukhov:
Hello! I'm opening this PR to start a discussion, so nevermind the lack of tests, I'm wiling to add them in case these changes make sense to the maintainers.

1. I'm not exactly sure why, but `request.protocol` is returning `http:` in our environment

    We're running the latest release versions of ember, ember-cli-fastboot, and fastboot-app-server.
    The node version is `6.11.1`. Also, here's relevant nginx config:

    ```
    location / {
      proxy_pass http://localhost:8081;
      proxy_set_header Host $host;
      proxy_set_header X-Real-IP $remote_addr;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header X-Forwarded-Proto $scheme;
    }
    ```

    Given the colon is appended in [ajax initializer](https://github.com/ember-fastboot/ember-cli-fastboot/blob/d7cb1b5766cf8057735cd9356b7a24ec57104b9e/fastboot/initializers/ajax.js#L9) and [documentation](https://github.com/ember-fastboot/ember-cli-fastboot#protocol) states the protocol is either `http` or `https` I assume this is a bug which needs to be addressed.

2. I believe fastboot should support `X-Forwarded-Proto` header

    The reason for that fastboot tries to do the best job of determining where to send API requests and protocol is an important part of that.

    In our use case, the fastboot/ember site _and_ API is accessible only with `https`. Given the fastboot server is running on `http` with nginx in front of it this presents a problem since API requests are also being sent over http. I know that in some simpler use cases you can specify the `host` in the adapter, but our situation is little tricky because the app is multi-tenant and it's served from arbitrary subdomains, and we want API requests to go to these exact subdomains. So supporting `X-Forwarded-Proto` seems to be the only way.

Please let me know if this makes sense. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,7 +276,7 @@ export default Ember.Route.extend({
 
 ### Protocol
 
-You can access the protocol (`http` or `https`) of the request that the
+You can access the protocol (`http:` or `https:`) of the request that the
 current FastBoot server is responding to via `fastboot.request` in the
 `fastboot` service.
 
--- a/fastboot/initializers/ajax.js
+++ b/fastboot/initializers/ajax.js
@@ -6,7 +6,7 @@ const { get } = Ember;
 var nodeAjax = function(options) {
   let httpRegex = /^https?:\/\//;
   let protocolRelativeRegex = /^\/\//;
-  let protocol = get(this, 'fastboot.request.protocol') + ':';
+  let protocol = get(this, 'fastboot.request.protocol');
 
   if (protocolRelativeRegex.test(options.url)) {
     options.url = protocol + options.url;
--- a/test/request-details-test.js
+++ b/test/request-details-test.js
@@ -49,8 +49,8 @@ describe('request details', function() {
       }
     })
       .then(function(response) {
-        expect(response.body).to.contain('Protocol: http');
-        expect(response.body).to.contain('Protocol from Instance Initializer: http');
+        expect(response.body).to.contain('Protocol: http:');
+        expect(response.body).to.contain('Protocol from Instance Initializer: http:');
       });
   });
 


kellyselden:
Which version of ember-cli-fastboot are you seeing this in. Did you see the recent PR #450 that changed how protocol works? Does that affect your findings at all?

krasnoukhov:
Well it's still reproducible on master branch

kratiahuja:
Please see https://github.com/ember-fastboot/fastboot/pull/119 for details why we added `:`

krasnoukhov:
Oh I see now. It makes sense, but I guess the fix now would be to remove second colon from [ajax initializer](https://github.com/ember-fastboot/ember-cli-fastboot/blob/d7cb1b5766cf8057735cd9356b7a24ec57104b9e/fastboot/initializers/ajax.js#L9)

kratiahuja:
Aah yes. I think so. Can you create a failing test?

krasnoukhov:
@kratiahuja Of course! What do you think about the `X-Forwarded-Proto` header though?

krasnoukhov:
@kratiahuja I pushed a fix/test/doc change for the double colon thing. It seems like `fastboot` repo would be more appropriate place to talk about `X-Forwarded-Proto` header, do you want me to open an issue or PR there? I'm wiling to help!

kratiahuja:
I am not convinced on the `X-Forwarded-Proto` yet. You can always have fastboot-server transition to https using nginx so that your API server also gets https:  https://github.com/ember-fastboot/ember-cli-fastboot/issues/379#issuecomment-296653467

krasnoukhov:
@kratiahuja I'm not sure what you suggest exactly. The comment you mentioned suggests to add `X-Forwarded-Proto` exactly like I did, but fastboot does not take that in account and thinks the original request was under `http`, which makes all ember-data API requests go to `http` as well. Does it make any sense?

kratiahuja:
@krasnoukhov I am suggesting your can have your `fastboot-app-server` run under `https` using the `x-forwarded-proto` header via a `beforeMiddleware` hook. There is really no need to leak those details in `fastboot` since this is amore of per client configuration. See this: https://github.com/ember-fastboot/fastboot-app-server/issues/36

krasnoukhov:
@kratiahuja Ok thanks for the link. Makes sense to me. Let me know if I can help more with double colon

kratiahuja:
Sure I'll merge this and release in 2 hours. Thanks for the fix!

kratiahuja:
ember-cli-fastboot@1.0.1 is published.

krasnoukhov:
Thank you @kratiahuja <3

PR #459, opened 2017-07-28
Simplify middleware used in tests to parse the body

cibernox:
In preparation of the docs, I've simplified it so it's more idiomatic.


test/fixtures/request/lib/post-middleware/index.js
```diff
@@ -1,33 +1,20 @@
...
-
-        var fastbootMiddleware = FastBootExpressMiddleware({
-          fastboot: this.fastboot
-        });
-
-        fastbootMiddleware(req, resp, next);
-      } else {
-        next();
-      }
+      fastbootMiddleware(req, resp, next);
```
kratiahuja: Shouldn't this middleware only be called when it is a POST request and when it is a base page request? if you have tests that request for assets, this would result in an error.

cibernox: I think we don't need to check for the method. `POST`, `PUT` and `PATCH` all can have body. `GET` will not, but it doesn't harm to leave it, right?

About the "base page request" what do you mean? That the request has a header requesting for HTML?

kratiahuja: > GET will not, but it doesn't harm to leave it, right?

yes

> About the "base page request" what do you mean? That the request has a header requesting for HTML?

I meant fastboot is serving the index.html page to you with the rendered DOM in it. You only want to send the request to the fastboot-express-middleware if it is a index.html request. You know it is because ember-cli exposes it via req.serveUrl being defined.

cibernox: > You only want to send the request to the fastboot-express-middleware if it is a index.html

Will that still work if the user is requesting `/some/nested/url`?

kratiahuja: yes if you are passing the accept/html request header.

cibernox: @kratiahuja I tried but I can't check for `req.serverUrl` yet until we fix [this line](https://github.com/ember-cli/ember-cli/blob/e38940e166435ebbc1ebbdc8ee49405b8b978946/lib/tasks/server/middleware/history-support/index.js#L58) in ember-cli

cibernox: I'll work on a PR to ember-cli itself right now.

kratiahuja:
@cibernox do you want me to wait on this PR or release?

cibernox:
Release if you can. This is just a simplification on the tests. Purely for my own satisfaction.

kratiahuja:
@cibernox 1.0.2 is released.

PR #460, opened 2017-07-29
set default port in readme

egaba88:
port `4200` is default, isnt it?


kratiahuja:
Thanks! We missed this.

PR #467, opened 2017-08-04
Fix for being used as nested addon

### LINKED ISSUE #466: `this.app` undefined when used from an addon
`ember-cli-staticboot` demonstrate this issue with latest `ember-cli-fastboot` https://github.com/robwebdev/ember-cli-staticboot/issues/30

Here is the code where it is happening:
https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L166

I am not familiar enough with addon internals to realise if `ember-cli-staticboot` is doing something that it should not or if this is just a use case that was not expected.

simonihmig:
Uses private(!) `this._findHost` or a shim if not available (ember-cli < 2.7) to get access to app, rather than parent addon (if nested).

Fixes #466

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,6 +18,17 @@ const Funnel = require('broccoli-funnel');
 const p = require('ember-cli-preprocess-registry/preprocessors');
 const existsSync = require('exists-sync');
 
+// Backport of _findHost for ember-cli < 2.7
+// See https://github.com/ember-cli/ember-cli/blob/0d17e839ca7a5128d1b95b9981a7e77ed0e55ead/lib/models/addon.js#L615
+function findHostShim() {
+  let current = this;
+  let app;
+  do {
+    app = current.app || app;
+  } while (current.parent.parent && (current = current.parent));
+  return app;
+}
+
 /*
  * Main entrypoint for the Ember CLI addon.
  */
@@ -42,7 +53,11 @@ module.exports = {
    *
    * See: https://ember-cli.com/user-guide/#integration
    */
-  included(app) {
+  included() {
+    let findHost = this._findHost || findHostShim;
+    let app = findHost.call(this);
+    this.app = app;
+
     // set autoRun to false since we will conditionally include creating app when app files
     // is eval'd in app-boot
     app.options.autoRun = false;


simonihmig:
@tchak would you be able to check if this change fixes your problem in #466?

tchak:
@simonihmig it fixes the problem, thanks!

rwjblue:
I'm unsure if we actually intend for ember-cli-fastboot to be a non-top-level dependency?

index.js
```diff
@@ -18,6 +18,17 @@ const Funnel = require('broccoli-funnel');
 const p = require('ember-cli-preprocess-registry/preprocessors');
 const existsSync = require('exists-sync');
 
+// Backport of _findHost for ember-cli < 2.7
+// See https://github.com/ember-cli/ember-cli/blob/0d17e839ca7a5128d1b95b9981a7e77ed0e55ead/lib/models/addon.js#L615
+function findHostShim() {
```
rwjblue: Do we need this?  Do we actually support ember-cli < 2.8?

simonihmig: Don't know. Only thing I found mentioned was this: `It is also preferable that your app is running ember-cli 2.12.0 and higher.` If this is a requirement it should be stated so...

Other problem with skipping this is that we are relying on a private API here (`_findHost`). Maybe this use case is a good reason to revisit https://github.com/ember-cli/ember-cli/pull/6755?

tchak:
@rwjblue do you have an idea on how to make work something like `ember-cli-staticboot` without this? https://github.com/robwebdev/ember-cli-staticboot/issues/30

kratiahuja:
I am not sure I understand why does `ember-cli-fastboot` need to be part of an addon? It's not meant for being part of addons.

kratiahuja:
After having looked at `ember-cli-staticboot`, I **strongly think** it doesn't need to have this addon as a dependnecy. It can dependend on libraries like `fastboot` to do the rendering and store the pages as static.

simonihmig:
I don't know enough about `ember-cli-staticboot`, but if you think it does not need `ember-cli-fastboot`, and there are no other good use cases for having it as a dependency, I am happy to close this.

kratiahuja:
@simonihmig I don't see a reason why addons need to have this addon as a dependency. It's fine to have it as a devDependency. I also don't know the history behind why `ember-cli-staticboot` has this addon as a dep. I would like understand that and then make a call.

From what I saw, I don't see a reason but maybe I am missing something. Maybe the addon author knows?

tchak:
I will investigate if I can make it work without depending on ´ember-cli-fastboot'

simonihmig:
Also worth mentioning: should you really need it, you can also add it to the app as a direct dependency, by calling [addAddonToProject](https://ember-cli.com/api/classes/Blueprint.html#method_addAddonToProject) in your addon's default blueprint. 

I think that is often the better solution to include a 3rd party addon than through a dependency. (user has control over addon's version, blueprints and custom commands work, which I *think* they don't when the addon is not a top-level dependency)

kratiahuja:
Going to close this since @tchak mentioned he is looking at ember-cli-staticboot to not depend on this addon.

nathanhammond:
The closing of this issue doesn't account for the case where you have an addon that supplies fastboot-related functionality and would like to use the embedded dummy application for testing.

In that case, `.app` is *also* `undefined` but `ember-cli-fastboot` is marked as a dev dependency and a (useless) peer dependency. `_findHost` is probably the wrong answer since we really want the behavior of something like `findRoot` in the sense that it is only intended to ever be the top item in the hierarchy (not engines, not other addons).

It's also possible that we instead simply detect the failure mode I describe and special case it.

PR #480, opened 2017-08-10
update ember-cli-babel

kellyselden:
This may help the ember-cli-htmlbars-inline-precompile upgrade.

cc @rwjblue


rwjblue:
👍 thanks @kellyselden

PR #485, opened 2017-08-13
remove compression

kellyselden:
This was used with the old command.


kratiahuja:
Thanks @kellyselden !

PR #490, opened 2017-08-13
tie test versions to package.json versions

kellyselden:
So they can drift via greenkeeper and stay in sync.


kratiahuja:
cc: @cibernox

cibernox:
Good idea. I suppose a new version came out and this broke.

Issue #502, opened 2017-08-16
Ember-CLI 2.15.0-beta.1 and greater pass autoboot options to browser JS


mixonic:
**tl;dr** Ember 2.15+ apps do not boot client-side application code with ember-cli-fastboot.

With Ember-CLI 2.14, ember-cli-fastboot will generate `{{app-boot}}` content like the following:

```js
if (typeof FastBoot === 'undefined') {
  if (!runningTests) {
    require('created/app')['default'].create({"name":"created","version":"0.0.0+60164923"});
  }
}

define('~fastboot/app-factory', ['created/app', 'created/config/environment'], function(App, config) {
  App = App['default'];
  config = config['default'];

  return {
    'default': function() {
      return App.create(config.APP);
    }
  };
});
```

In Ember-CLI 2.15.0-beta.1 and greater, the `{{CONFIG_APP}}` value that becomes arguments to `create` in the browser ([source](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/lib/utilities/fastboot-app-module.js#L15)) now includes `"autoboot": false`:

```js
if (typeof FastBoot === 'undefined') {
  if (!runningTests) {
    require('created/app')['default'].create({"name":"created","version":"0.0.0+60164923","autoboot":false});
  }
}

define('~fastboot/app-factory', ['created/app', 'created/config/environment'], function(App, config) {
  App = App['default'];
  config = config['default'];

  return {
    'default': function() {
      return App.create(config.APP);
    }
  };
});
```

This prevents the browser from booting the app. It isn't completely clear if Ember-CLI fixed a bad bug ember-cli-fastboot was reliant upon, if Ember-CLI introduced a regression, or if there is something less than optimal on both sides.

/cc @kratiahuja @rwjblue

rwjblue:
Thank you for reporting with such detail. 

It seems this is caused by a recent addition of caching the output of `project.confjg()` calls. Previously, every time you call `project.config()` you would get a fresh object and could therefore treat it as your own "copy". We determined that building the config is actually fairly expensive so we added caching, which ultimately meant that this mutation we were doing now modifies the single shared config. 

@kratiahuja has a fix up in #503 for this.

mixonic:
Closing this in favor of https://github.com/ember-fastboot/ember-cli-fastboot/pull/503, which I can confirm fixes the issue. Thank you!

PR #503, opened 2017-08-16
Set autoBoot to false only for fastboot config

kratiahuja:
Fixes https://github.com/ember-fastboot/ember-cli-fastboot/issues/502

When setting the config for `fastboot`, we want to set `autoBoot` to false _only_ for fastboot config, so that the app doesn't route automatically and render a template. Routing and rendering is done on per request basis via the `visit` API.

It seems this bug has only showed up since `config` value seems to be cached across Ember CLI hooks. Still digging into what changed in Ember CLI 2.15+ to cause this.

Tested manually with an app using Ember CLI 2.15. Need to update the version of Ember CLI in fastboot addon so tests run against that version too.

cc: @mixonic @rwjblue @stefanpenner


kratiahuja:
Tests are failing on an unrelated error. Will dig into it.

kratiahuja:
Ok tests should _hopefully_ pass now. Didn't realize `JSON.stringy(/foo/)` or `JSON.stringify(new RegExp(/foo/))` gives `{}` so tests were failing.

kratiahuja:
@rwjblue would love your eyes again on this since I had to hand roll the clone functionality.

kratiahuja:
Yup agreed.

PR #506, opened 2017-08-16
add ember-cli-preprocess-registry to deps

kellyselden:
Used https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L18

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "broccoli-plugin": "^1.2.1",
     "chalk": "^2.0.1",
     "ember-cli-babel": "^6.7.2",
+    "ember-cli-preprocess-registry": "^3.1.1",
     "ember-cli-version-checker": "^2.0.0",
     "fastboot": "^1.0.0",
     "fastboot-express-middleware": "^1.0.0",


rwjblue:
Not sure we should actually add the dep here.  We need to require ember-cli's own version of this or we are going to cause many issues down the line when we update it in ember-cli.

kellyselden:
@rwjblue Does ember-cli re-export it that you know of? This was causing failures when I tried to npm link ember-cli.

kratiahuja:
Isn't `ember-cli-preprocess-registry` standalone? I thought `ember-cli` depends on it as well.

rwjblue:
It does, I've just been working on some breaking changes (for ember-cli-preprocess-registry@4) and dealing with the interactions between the two different versions (one from  ember-cli-fastboot and one from ember-cli) will be difficult...

kellyselden:
@rwjblue We could solve this via a simple reexport in ember-cli right?

kratiahuja:
I am fine with one source of truth (even if it means `ember-cli`).

PR #508, opened 2017-08-16
clean up shoebox test code

kellyselden:


kratiahuja:
I would prefer to delet from tests/index.html

PR #515, opened 2017-08-27
Expose fastbootConfigTree hook

### LINKED ISSUE #513: Expose `fastbootConfigTree` hook from this addon
This issue tracks the fastboot work for enabling engines support in FastBoot.

# Motivation

With Ember engines, each lazy engine provides it own configuration object just like the host app. In order for engines to work in FastBoot in a performant way, we need this configuration to be available in FastBoot sandbox as well. 

# Public API

FastBoot addon will recognize a `fastbootConfigTree` API like `updateFastBootManifest` which will be responsible for walking through the nested addons and invoke `fastbootConfigTree` on each defined addon. The `fastbootConfigTree` will return a map of config names to the configuration.

FastBoot addon will collect and merge these maps into one and write it to package.json.

# How does engine get it's configuration in sandbox?

FastBoot already has a public method called `FastBoot.config()` in the sandbox which returns the app configuration for the host app as a JSON object. We need to extend this method to take an _optional_ key. Engine will provide the key when invoking `FastBoot.config` method. If the key isn't provided, we will default to returning the host app configuration in FastBoot.

cc: @rwjblue

kratiahuja:
Fixes #513 . This unblocks engines support in FastBoot

Exposes `fastBootConfig` tree hook that addons can use to write additional config to expose in Node. In addition also bumped the schema version since the config schema is changed a bit. 

# TODO:
- [x] Merge and release : https://github.com/ember-fastboot/fastboot/pull/164
- [x] Update fastboot version
- [ ] Update fastboot website with additional hook

cc: @rwjblue

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,25 @@ treeForFastBoot: function(tree) {
 
 The `tree` is the additional fastboot asset that gets generated and contains the fastboot overrides.
 
+### Providing additional config
+
+By default `ember-cli-fastboot` reads the app's config and provides it in the FastBoot sandbox as a JSON object. For the app in browser, it respects `storeConfigInMeta` and either reads it from the config meta tag or inlines it as JSON object in the `app-name/config/environment` AMD module.
+
+Addons like ember-engines may split the app in different bundles that are loaded asynchronously. Since each bundle is loaded asynchronously, it can have its own configuration as well. In order to allow FastBoot to provide this config in the sandbox, it exposes a `fastbootConfigTree` build hook.
+
+Addons wishing to use this hook simply need to return a unique identifier for the configuration with the configuration.
+
+```js
+fastbootConfigTree() {
+  return {
+    '<engine-name>': {
+      'foo': 'bar'
+    }
+  }
+}
+```
+
+The above configuration will be available in Node via the `FastBoot.config()` function. Therefore, in order to get the above config, the addon/app can call `FastBoot.config('<engine-name>')`.
 
 ## Known Limitations
 
--- a/index.js
+++ b/index.js
@@ -84,11 +84,12 @@ module.exports = {
     // if the fastboot addon is installed, we overwrite the config-module so that the config can be read
     // from meta tag/directly for browser build and from Fastboot config for fastboot target.
     if (type === 'config-module') {
-      var originalContents = contents.join('');
+      const originalContents = contents.join('');
+      const appConfigModule = `${config.modulePrefix}`;
       contents.splice(0, contents.length);
       contents.push(
         'if (typeof FastBoot !== \'undefined\') {',
-          'return FastBoot.config();',
+          'return FastBoot.config(\'' + appConfigModule + '\');',
         '} else {',
           originalContents,
         '}'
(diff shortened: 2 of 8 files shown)


kratiahuja:
Tests will pass once new version of `fastboot` is published and updated.

package-lock.json
```diff
@@ -0,0 +1,11312 @@
+{
```
rwjblue: We should stick with either `yarn.lock` or `package-lock.json` in this repo.

index.js
```diff
@@ -84,11 +84,12 @@ module.exports = {
...
     // from meta tag/directly for browser build and from Fastboot config for fastboot target.
     if (type === 'config-module') {
-      var originalContents = contents.join('');
+      const originalContents = contents.join('');
+      const appConfigModule = `${config.modulePrefix}/config/environment`;
       contents.splice(0, contents.length);
       contents.push(
         'if (typeof FastBoot !== \'undefined\') {',
-          'return FastBoot.config();',
+          'return FastBoot.config(\'' + appConfigModule + '\');',
```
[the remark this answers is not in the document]
kratiahuja: No it doesn't have to change. Just wanted to make it consistent and less magic :)

package-lock.json
kratiahuja: Yes I didn't mean to commit package-lock.json.

lib/broccoli/fastboot-config.js
```diff
@@ -78,6 +82,23 @@ module.exports = class FastBootConfig extends Plugin {
     }
   }
 
+  buildConfig() {
+    // we only walk the host app's addons to grab the config since ideally
+    // addons that have dependency on other addons would never define
+    // this advance hook.
+    this.project.addons.forEach((addon) => {
+      if (addon.fastbootConfigTree) {
+        let configFromAddon = addon.fastbootConfigTree();
```
SergeAstapov: @kratiahuja should `environment` be passed to this hook as an argument? Similar to how `environment ` is provided to ember-cli Addon `config` hook https://ember-cli.com/api/classes/Addon.html#method_config

kratiahuja: No I don't see that is something going to be used by addons unless you have a usecase in mind.

kratiahuja:
Ok so all tests on Node 6 and Nod 8 should pass now. Node 4 will pass once [ember-data issue is merged](https://github.com/emberjs/data/pull/5187).

@rwjblue this should be ready for another review.

README.md
```diff
@@ -432,6 +432,25 @@ treeForFastBoot: function(tree) {
...
+By default `ember-cli-fastboot` reads the app's config and provides it in the FastBoot sandbox as a JSON object. For the app in browser, it respects `storeConfigInMeta` and either reads it from the config meta tag or inlines it as JSON object in the `app-name/config/environment` AMD module.
+
+Addons like ember-engines may split the app in different bundles that are loaded asynchronously. Since each bundle is loaded asynchronously, it can have its own configuration as well. In order to allow FastBoot to provide this config in the sandbox, it exposes a `fastbootConfigTree` build hook. 
+
+Addons wishing to use this hook simply need to return a unique identifier for the configuration with the configuration.
+
+```js
+fastbootConfigTree() {
+  return {
+    '<engine-name>/config/environment': {
```
rwjblue: I think that the key here now should be `<engine-name>`.

kratiahuja:
Node 4 is breaking again on ember data.

kratiahuja:
Restarted Node4 travis job. If it passes, I'll merge and release.

kratiahuja:
Thank you @SergeAstapov for the help here!

PR #525, opened 2017-09-11
[WIP] Add AMP HTML File to Manifest

devotox:


kratiahuja:
This approach makes it very specific rather than generic. I believe you could achieve the same thing using `beforeMiddleware` hook of `fastboot-app-server`.

kratiahuja:
Closing per discussion in the other PR.

PR #540, opened 2017-10-18
add ember-cached-shoe info to README.md

HarenBroog:


README.md
```diff
@@ -376,6 +376,27 @@ export default Ember.Route.extend({
 });
 ```
 
+### Think out of the Shoebox
+
+Shoebox gives you great capabilities, but using in in the real app is pretty rough. Have you ever thought that such kind of logic should be done behind the scenes? In a large codebase, defining `fastboot.isFastboot` conditionals can be a daunting task. Furthermore, it generates a lot of boilerplate code, which obscures the solution. Sooner or later coupling with `shoebox` will spread over all routes.
```
SergeAstapov: @HarenBroog typo: `using in in the real app` should be `using **it** in the real app`?

README.md
```diff
@@ -376,6 +376,27 @@ export default Ember.Route.extend({
...
+After installing and applying it, your routes can look like this:
+
+`app/routes/my-route.js`:
+
+```javascript
+import Ember from 'ember'
+
+export default  Ember.Route.extend({
+  model() {
+    this.store.findAll('posts') // called twice, second call serves cached response
```
SergeAstapov: @HarenBroog `called twice` seems to be kinda confusing. What if it would say something like
```
first call in server makes actual ajax request.
second call in browser serves cached response
```

Also minor things - example is missing `return` statement.

HarenBroog: Thanks for your comments mate, both of them make sense :) 👍

PR #544, opened 2017-10-25
Allow to extend application FastBoot config via addon hook fastbootConfigTree

SergeAstapov:
Use lodash.merge provided by `ember-cli-lodash-subset` instead of Object.assign for deep merge

Added new app fixture `fastboot-config` that copied from `module-whitelist`.
Implemented `fastbootConfigTree` hook in `test/fixtures/fastboot-config/node_modules/fake-addon/index.js` that extends application config to check for a regression and fix actually works.

### DIFF
--- a/test/package-json-test.js
+++ b/test/package-json-test.js
@@ -84,6 +84,12 @@ describe('generating package.json', function() {
       ]);
     });
 
+    it('contains app name', function() {
+      let pkg = fs.readJsonSync(app.filePath('dist/package.json'));
+
+      expect(pkg.fastboot.appName).to.equal('module-whitelist');
+    });
+
     it('contains the application config', function() {
       let pkg = fs.readJsonSync(app.filePath('dist/package.json'));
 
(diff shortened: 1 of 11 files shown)


SergeAstapov:
I'm I missing anything so that tests are failing only on Travis only on node@4?

test/package-json-test.js
```diff
@@ -99,15 +99,50 @@ describe('generating package.json', function() {
...
     });
 
-    it('contains additional config from addons', function() {
+    it('contains additional config from ember-fastboot-build-example addon', function() {
       let pkg = fs.readJsonSync(app.filePath('dist/package.json'));
 
       expect(pkg.fastboot.config['foo']).to.equal('bar');
     });
 
+    describe('with addon that implements fastbootConfigTree', function() {
```
kratiahuja: Would recommend moving this after the `'contains app name` test so it is easier to read.

SergeAstapov: @kratiahuja I've moved `'contains app name'` up to make tests for application config grouped.
Thanks for feedback!

SergeAstapov:
@kratiahuja Could you please take a look if updated tests order makes sense to you?

PR #546, opened 2017-11-03
Disentangle from broccoli-asset-rev

ef4:
ember-cli-fastboot has a lot of intimate knowledge of broccoli-asset-rev, and it's forced to always run after broccoli-asset-rev.

This increases complexity and reduces flexibility, resulting in problems like https://github.com/ef4/prember/issues/2, in which we want to use the output of ember-cli-fastboot to render some content HTML which will then need to run through broccoli-asset-rev.

Rewriting assets is broccoli-asset-rev's responsibility, so we can just teach it how to properly rewrite a fastboot-capable ember app. I've done that work [here](https://github.com/ef4/broccoli-asset-rev#838f1b7186fde566d3e8010196c1f974423a4946).

This PR is the corresponding change in ember-cli-fastboot, which mostly just rips out unneeded code.

This is not an upgrade that can be done gracefully, however, since addons cannot dynamically change their `before` and `after` declarations, and my entire motiviation in doing this is to get rid of `after: ['broccoli-asset-rev']`. So this PR creates a hard requirement that the user has a correspondingly new broccoli-asset-rev. I added a clear error message so that it should be only a small stumbling block.

This PR will not pass the test suite, because ember-cli-addon-tests is apparently incapable of being told what dependencies to use. Perhaps it would be best to simply wait to see if we can get the newer broccoli-asset-rev released and upgraded in the default app blueprint before merging this.

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,6 +12,7 @@ const chalk = require('chalk');
 const fastbootAppModule = require('./lib/utilities/fastboot-app-module');
 const FastBootConfig = require('./lib/broccoli/fastboot-config');
 const migrateInitializers = require('./lib/build-utilities/migrate-initializers');
+const SilentError = require('silent-error');
 
 const Concat = require('broccoli-concat');
 const Funnel = require('broccoli-funnel');
@@ -48,17 +49,18 @@ module.exports = {
    * See: https://ember-cli.com/user-guide/#integration
    */
   included(app) {
+
+    let assetRev = this.project.addons.find(addon => addon.name === 'broccoli-asset-rev');
+    if(assetRev && !assetRev.supportsFastboot) {
+      throw new SilentError("This version of ember-cli-fastboot requires a newer version of broccoli-asset-rev");
+    }
+
     // set autoRun to false since we will conditionally include creating app when app files
     // is eval'd in app-boot
     app.options.autoRun = false;
 
-    if (app.options.fingerprint) {
-      // set generateAssetMap to be true so that manifest files can be correctly written
-      // in package.json
-      app.options.fingerprint.generateAssetMap = true;
-    }
-
     app.import('vendor/experimental-render-mode-rehydrate.js');
+
     // get the app registry object and app name so that we can build the fastboot
     // tree
     this._appRegistry = app.registry;
@@ -285,7 +287,6 @@ module.exports = {
     let fastbootAppConfig = appConfig.fastboot;
 
     return new FastBootConfig(tree, {
-      assetMapPath: this.assetMapPath,
       project: this.project,
       name: this.app.name,
       outputPaths: this.app.options.outputPaths,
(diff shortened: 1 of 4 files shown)


ef4:
Linked PR https://github.com/rickharrison/broccoli-asset-rev/pull/125

SergeAstapov:
@ef4 for the tests sake you can modify `package.json` for each test, see for example https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/test/package-json-test.js#L243

I believe you can locally point `broccoli-asset-rev` to your fork, run tests and modify if needed.
This does not need to be committed but you can ensure tests are ok. And this PR build can be re-run after `broccoli-asset-rev` is merged to check that everything looks good if this is approved by maintainers.

Just my two cents about tests.

BTW, good idea 👍

ef4:
Thanks @SergeAstapov.

Getting the whole test suite to use an additional dependency requires significantly refactoring most of the tests, since only a few are already using `skipNpm: true` with their own later `npm install`. Without that, `editPackageJSON` doesn't actually get you an additional module.

I was able to see the whole test suite passing through some abuse of test:precook, plus edits to the test modules that are already doing their own npm installs.

rwwagner90:
Definitely would love to see this stuff merged! Hitting some cycle issues with prember and ember-service-worker.

ef4:
We can't land this PR until there is a decision on https://github.com/rickharrison/broccoli-asset-rev/pull/125

ef4:
The broccoli-asset-rev pr is merged and released, and this is now green and ready to go.

rwjblue:
Hmm, as written this is a breaking change. Can we make the current error a warning, and do the old behavior when broccoli-asset-rev is too old?

ef4:
The problem is the old behavior includes depending on `after` in package.json. We would need to dynamically change the ordering, which I don't think is a thing in ember-cli. And we don't actually fix the bug unless we get rid of `after` in the cases where we're using the new feature.

ef4:
@rwjblue can we accept this as a breaking change, or do we need to invent a hack to do dynamic plugin ordering in ember-cli?

rwjblue:
TBH, I think we should just land and release 2.0.0. I don't think the work around hacks are very viable, and there really isn't a good reason not to just bump majors here IMHO.

@ef4 / @kratiahuja / @stefanpenner / @tomdale / @kellyselden - Any objections?

kellyselden:
I am no longer afraid of major versions. I say bump away.

ef4:
I hear consensus for merging this and bumping the major.

rwwagner90:
+1

rwwagner90:
We really need this for prember. Any chance we can get this merged and bump the major version? If there are more issues/PRs that need to be resolved first, I am happy to help get us there, just let me know how I can help!

ef4:
Now we just need somebody with NPM auth to cut the release.

rwjblue:
ember-cli-fastboot 2.0.0 published 🎉

PR #557, opened 2017-11-29
update list of hooks called by fastboot

nlfurniss:
Spoke with @kratiahuja and these hooks need to be added.


nlfurniss:
@kratiahuja can we get this in?

PR #566, opened 2018-01-14
Update to Ember & Ember-cli 2.18.0

cibernox:


app/locations/none.js
```diff
@@ -1,16 +1,12 @@
-import Ember from 'ember';
-
-const {
-  computed,
-  computed: { bool, readOnly },
-  inject: { service },
-  get,
-  getOwner
-} = Ember;
+import { computed, get } from '@ember/object';
```
rwjblue: I don't believe we can make these changes in `app/`. Sadly the codemod automatically does both `app` and `addon` directories, but within an addon we can only be certain that our own `addon/` (and `addon-test-support/`) are transpiled by a new enough ember-cli-babel version...

rwjblue: We should probably move this code into `addon/` then re-export in `app`...

app/services/fastboot.js
```diff
@@ -1,10 +1,12 @@
 /* global FastBoot */
-import Ember from "ember";
-
-const { deprecate, computed, get, assert } = Ember;
-const { deprecatingAlias, readOnly } = computed;
-
-const RequestObject = Ember.Object.extend({
+import { deprecate } from '@ember/application/deprecations';
```
rwjblue: Same as above, we should move this file to addon/ and reexport only in `app`

lib/broccoli/fastboot-config.js
```diff
@@ -143,11 +143,10 @@ module.exports = class FastBootConfig extends Plugin {
 
     this.dependencies = dependencies;
     this.moduleWhitelist = uniq(moduleWhitelist);
-  };
```
rwjblue: Umm, how did this work before?  Is this file even used? This seems like it would have been a parser error to me?!?!?

cibernox: It seems to only be a lint problem

test/package-json-test.js
```diff
@@ -99,7 +99,7 @@ describe('generating package.json', function() {
         baseURL: '/',
         locationType: 'auto',
         EmberENV: { FEATURES: {} },
-        APP: { name: 'module-whitelist', version: '0.0.0+', autoboot: false },
+        APP: { name: 'module-whitelist', version: '0.0.0', autoboot: false },
```
cibernox: @rwjblue This was the only failing test, but I have no clue why there was a plus sign in this version number.

kratiahuja: Yeah the current master is red due to these tests failing. Thanks for fixing this.

addon/instance-initializers/clear-double-boot.js
```diff
@@ -0,0 +1,28 @@
...
+
+export default {
+  name: "clear-double-boot",
+
+  initialize: function(instance) {
+    if (typeof FastBoot === 'undefined') {
+      var originalDidCreateRootView = instance.didCreateRootView;
+
+      instance.didCreateRootView = function() {
+        let elements = document.querySelectorAll(instance.rootElement + ' .ember-view');
```
kratiahuja: When can we have multiple ember-views?

cibernox: I just moved the code around. I also don't see how this is possible either, but I wouldn't worry because I'm about to change how we mark the beginning and end of the app so fastboot works when the experiemental flag `ember-glimmer-remove-application-template-wrapper` is enabled.
We'll probably need some script tag markets to delimitate the boundaries.

Issue #563, opened 2018-01-02
Ember CLI 3.0 compatibility


mixonic:
ember-cli-fastboot is not compatible with Ember CLI 3.0-beta.1. Known issues:

* [ ] In https://github.com/ember-cli/ember-cli/pull/7525 Ember CLI removed `project.nodeModulesPath`. This is used in ember-cli-fastboot for [detection of app initializers provided by addons](https://github.com/ember-fastboot/ember-cli-fastboot/blob/fb410e1f32b1b70a512de8b0a8742570eb672962/lib/build-utilities/migrate-initializers.js#L92).

rwjblue:
Thanks for reporting!

rondale-sc:
I ran into this one as well.

kratiahuja:
I'll look into this next week. Apologies for the delay here. Have been on vacation.

0xadada:
Just tried upgrading to Ember 3.0.0.beta.1 using `ember update` command from ember-cli-update addon. After resolving conflicts, i ran `ember s` and ran into the error below.

Just ran into this problem as well, initially filed over at https://github.com/ember-cli/ember-cli/issues/7552

upgrade is on a branch here https://github.com/mirai-audio/mir/tree/16/library-ember-3beta

---

Output from `ember version --verbose && npm --version && yarn --version`:
```
$ ember version --verbose && npm --version && yarn --version
ember-cli: 3.0.0-beta.1
http_parser: 2.7.0
node: 8.9.3
v8: 6.1.534.48
uv: 1.15.0
zlib: 1.2.11
ares: 1.10.1-DEV
modules: 57
nghttp2: 1.25.0
openssl: 1.0.2n
icu: 59.1
unicode: 9.0
cldr: 31.0.1
tz: 2017b
os: darwin x64
5.5.1
1.3.2
```


trace of bug
```
$ ember s
Path must be a string. Received undefined


Stack Trace and Error Report: /var/folders/ss/sjsdtptn7j1cc6y_x8lbc3ch0000gn/T/error.dump.118f766ba9e6854a07cd61a8428bd49e.log
```
then
```
$ cat /var/folders/ss/sjsdtptn7j1cc6y_x8lbc3ch0000gn/T/error.dump.118f766ba9e6854a07cd61a8428bd49e.log
=================================================================================

ENV Summary:

  TIME: Fri Jan 12 2018 16:30:30 GMT-0500 (EST)
  TITLE: ember
[...]

PR #567, opened 2018-01-15
Fix for #563 Ember 3.0 compatability issue

0xadada:
Implemented backwards compatibility fix

Fixes https://github.com/ember-fastboot/ember-cli-fastboot/issues/563

### DIFF
--- a/lib/build-utilities/migrate-initializers.js
+++ b/lib/build-utilities/migrate-initializers.js
@@ -89,10 +89,8 @@ function _moveFastBootInitializers(project, rootPath) {
  * @param {Object} project
  */
 function _migrateAddonInitializers(project) {
-  const nodeModulesPath = path.join(project.nodeModulesPath);
-
-  project.addons.forEach((addon) => {
-    const currentAddonPath = path.join(nodeModulesPath, addon.name);
+  project.addons.forEach(addon => {
+    const currentAddonPath = addon.root;
 
     _checkBrowserInitializers(currentAddonPath);
     _moveFastBootInitializers(project, currentAddonPath);


lib/build-utilities/migrate-initializers.js
```diff
@@ -89,10 +89,17 @@ function _moveFastBootInitializers(project, rootPath) {
...
-  const nodeModulesPath = path.join(project.nodeModulesPath);
-
-  project.addons.forEach((addon) => {
-    const currentAddonPath = path.join(nodeModulesPath, addon.name);
+  // `nodeModulesPath` was removed in Ember 3, but we still support older
+  // versions, use empty string as a fallback
+  const legacyNodeModulesPath = path.join(project.nodeModulesPath || '');
+
+  project.addons.forEach(addon => {
+    const currentAddonPath = addon.root
```
rwjblue: I think we can always rely on addon.root being the full path to the addons files. It has been around for a very long time (I think since nested addons were added back in 0.2.0).

Can you refactor to that alone (without fallback)?

0xadada: Sure thing

0xadada: @rwjblue all set 👍

lib/build-utilities/migrate-initializers.js
```diff
@@ -89,10 +89,8 @@ function _moveFastBootInitializers(project, rootPath) {
  * @param {Object} project
  */
 function _migrateAddonInitializers(project) {
-  const nodeModulesPath = path.join(project.nodeModulesPath);
-
-  project.addons.forEach((addon) => {
-    const currentAddonPath = path.join(nodeModulesPath, addon.name);
+  project.addons.forEach(addon => {
+    const currentAddonPath = path.join(addon.root);
```
rwjblue: I think we can make this:

```
const currentAddonPath = addon.root;
```

Specifically, it seems odd to call path.join with only one arg...

0xadada: @rwjblue agree, i just pushed a simplified version.

rwjblue:
Thank you!

Issue #565, opened 2018-01-12
Doesn't work when `ember-glimmer-remove-application-template-wrapper` flag is enabled.


cibernox:
Using Ember canary, if the feature flag `ember-glimmer-remove-application-template-wrapper` is enabled when the JS boots the SSR content is not replaced, ending on a repeated page:

<img width="690" alt="screen shot 2018-01-12 at 21 12 40" src="https://user-images.githubusercontent.com/265339/34893411-6196bd36-f7dd-11e7-88a7-08c63dddee16.png">

rwjblue:
Awesome, thanks for reporting!  The issue here is that [this initializer](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/app/instance-initializers/clear-double-boot.js#L18) can no longer clear the "prior boot"...

rwjblue:
We can change that to add another mechanism to track the boundaries.  For example, if we rendered the user content like this:

```html
<script type="x/boundary" id="fastboot-start"></script>

<!-- user rendered content goes here -->

<script type="x/boundary" id="fastboot-end"></script>
```

You could easily imagine implementing a bit of DOM traversal to remove all the "guts" between those markers....

cibernox:
@rwjblue I can see how to modify the initializers to remove stuff between two known nodes, but any idea how to change the rendering to add those boundary markers.

rwjblue:
Ya it’s in fastboot package

mike183:
I believe #569 may have caused the shoebox data to be stripped from the page.

Reason being that if you look at the page source that was actually returned from Fastboot, it contains the shoebox data, where as viewing the page source in the 'Elements' tab of Chrome's dev tools, the shoebox content appears to have disappeared.

Has anyone else experienced this?

I assume the fix would be a simple case of updating the instance initializer to ignore `script` tags with a type of `fastboot/shoebox`?

PR #569, opened 2018-01-18
Remove content between markers

cibernox:
Ready to merge


cibernox:
Ok, I've now verified that this works with and without the `ember-glimmer-remove-application-template-wrapper` flag enabled, and in both cases it leaves the DOM clean of any old content or markers.

The only thing I'm not sure how to do is ensure that the `clear-double-boot` initializer runs after the shoebox information has been loaded. I actually think that this should not be enforced in this addon but on the addons expecting to use the shoebox instead. It seems more correct for those addons to be aware how fastboot works than fastboot being aware of all addons using a shoebox-like approach.

cibernox:
@rwjblue @kratiahuja This is now ready 🎉

kratiahuja:
@cibernox can please we make sure to add a test to make sure we don't regress this?

cibernox:
I've added tests. I had to use imagination a bit to add a test for the initializer, but it's the best idea I came up with.

kratiahuja:
Thank you @cibernox

cibernox:
@kratiahuja Can I get a new release at some point this week?

kratiahuja:
@cibernox Sure I will release tomorrow.

cibernox:
@kratiahuja Did you forget about the release? 😉

kratiahuja:
@cibernox Super sorry! Yes I forgot about the release. v1.1.3 is released.

cibernox:
I've just verified that it works on an app using canary wth the flag
`ember-glimmer-remove-application-template-wrapper` enabled 🎉

2018-01-30 16:39 GMT+01:00 Krati Ahuja <notifications@github.com>:

Sweet-Bob:
After 1.1.3 update i had a problem like this https://github.com/ember-fastboot/ember-cli-fastboot/issues/565 (Ember 2.18.0), but only with fastboot-app-server.
On 1.1.2 works fine.

cibernox:
@Sweet-Bob Someone else reported it too. It's a problem in fastboot-app-server, it has to be upgraded to 1.1.3 too.

Sweet-Bob:
@cibernox Should be reported to fastboot-app-server repo?

cibernox:
PR done: https://github.com/ember-fastboot/fastboot-app-server/pull/73

Sweet-Bob:
@cibernox Thank you!

PR #578, opened 2018-02-23
BUGFIX: fastboot/shoebox tags should not be removed in clear-double-b…

snewcomer:
…oot instance initializer

Ref https://github.com/ember-fastboot/ember-cli-fastboot/pull/569/files

### DIFF
--- a/addon/instance-initializers/clear-double-boot.js
+++ b/addon/instance-initializers/clear-double-boot.js
@@ -6,15 +6,24 @@
 //
 // This removes any pre-rendered ember-view elements, so that the booting
 // application will replace the pre-rendered output
+//
+const START = 'fastboot-body-start';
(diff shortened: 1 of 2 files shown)


snewcomer:
@rwjblue @kratiahuja I "think" `clear-double-boot` was removing shoebox elements and such.  

Also is the intention to load off all shoebox data before those elements are cleared or is it intended that a user is able to grab stuff off the shoebox, say from a component?

snewcomer:
And since maybe the desire it to support a non root wrapping element, perhaps we need to ensure some configuration...i.e. the non root wrapping element should have a class of `.ember-view` or something.  Lmk if I am missing some context.

snewcomer:
I'm thinking I didn't understand a few things properly.  I'll close for now.

PR #580, opened 2018-02-23
Utilize rehydration serialization from glimmer

rondale-sc:
This is part of an arching plan to introduce Glimmer's
rehydration/serializtion modes to Ember proper.

There are 4 PR's that are all interwoven, of which, this is one.

- [x] Glimmer.js: glimmerjs/glimmer-vm#783 (comment)
- [x] Ember.js: emberjs/ember.js#16227
- [x] Fastboot: ember-fastboot/fastboot#185
- [ ] EmberCLI Fastboot: ember-fastboot/ember-cli-fastboot#580

In order of need to land they are:

Glimmer.js:
glimmerjs/glimmer-vm#783 (comment)

This resolves a rather intimate API problem.  Glimmer-vm expects a very
specific comment node to exist to know whether or not the rehydration
element builder can do it's job properly.  If it is not found on the
first node from `rootElement` it throws.

In fastboot however we are certain that there will already be existant
elements in the way that will happen before rendered content.

This PR just iterates through the nodes until it finds the expected
comment node.  And only throws if it never finds one.

---

Ember.js:
emberjs/ember.js#16227

This PR modifies the `visit` API to allow a private _renderMode to be
set to either serialize or rehydrate.  In SSR environments the
assumption (as it is here in this fastboot PR) is that we'll set the
visit API to serialize mode which ensures glimmer-vm's serialize element
builder is used to build the API.

The serialize element builder ensures that we have the necessary fidelty
to rehydrate correctly and is mandatory input for rehydration.

---

Fastboot:
ember-fastboot/fastboot#185

This allows enviroment variable to set _renderMode to be used in Visit
API.  Fastboot must send content to browser made with the serialization
element builder to ensure rehydration can be sucessful.

---

EmberCLI Fastboot:
ember-fastboot/ember-cli-fastboot#580

Finally this does the fun part of disabling the current
clear-double-render instance-initializer

We first check to ensure we are in a non-fastboot environment.  Then we
ensure that we can find the expected comment node from glimmer-vm's
serialize element builder.  This ensures that this change will only
effect peoeple who use ember-cli-fastboot with the serialized output
from the currently experimental fastboot setup

Then we ensure `ApplicationInstance#_bootSync` specifies the rehydrate
_renderMode.

This is done in `_bootSync` this way because visit is currently not used
to boot ember applications.  And we must instead set bootOptions this
way instead.

We also remove the markings for `fastboot-body-start` and
`fastboot-body-end` to ensure clear-double render instance-initializer
is never called.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,8 @@ it to know whether or not the DOM it received in the browser side was generated
 by the serialization builder.  If it was, it tells the Ember.js Application to
 use the rehydration builder and your application will be using rehydration.
 
+Rehydration is only compatible with fastboot > 1.1.4, and Ember.js > 3.2.
+
 ## Build Hooks for FastBoot
 
 ### Disabling incompatible dependencies
--- a/README.md
+++ b/README.md
@@ -434,7 +434,7 @@ it to know whether or not the DOM it received in the browser side was generated
 by the serialization builder.  If it was, it tells the Ember.js Application to
 use the rehydration builder and your application will be using rehydration.
 
-Rehydration is only compatible with fastboot > 1.1.4, and Ember.js > 3.2.
+Rehydration is only compatible with fastboot > 1.1.4-beta.1, and Ember.js > 3.2.
 
 ## Build Hooks for FastBoot
 
--- a/README.md
+++ b/README.md
@@ -399,6 +399,41 @@ export default  Ember.Route.extend({
 ```
 And they still take advantage of caching in the `shoebox`. No more redundant AJAX for already acquired data. Installation details are available in the addon [README](https://github.com/appchance/ember-cached-shoe#ember-cached-shoe).
 
+### Rehydration
+
+What is Rehydration?
+
+The rehydration feature means that the Glimmer VM can take a DOM tree
+created using Server Side Rendering (SSR) and use it as the starting
+point for the append pass.
+
+See details here:
+
+https://github.com/glimmerjs/glimmer-vm/commit/316805b9175e01698120b9566ec51c88d075026a
+
+In order to utilize rehydration in Ember.js applications we need to ensure that
+both server side renderers (like fastboot) properly encode the DOM they send to
+the browser with the serialization format (introduced in the commit above) AND
+that the browser instantiated Ember.js application knows to use the rehydration
+builder to consume that DOM.
+
+Rehydration is 100% opt-in, if you do not specify the environment flag your
+application will behave as it did before!
+
+We can opt-in to the rehydration filter by setting the following environment
+flag:
+
+```
+EXPERIMENTAL_RENDER_MODE_SERIALIZE=true
+```
+
+This flag is read by Ember CLI Fastboot's dependency; fastboot to alert it to
+produce DOM with the glimmer-vm's serialization element builder.  This addon
+(Ember CLI Fastboot) then uses a utility function from glimmer-vm that allows
+it to know whether or not the DOM it received in the browser side was generated
+by the serialization builder.  If it was, it tells the Ember.js Application to
+use the rehydration builder and your application will be using rehydration.
+
 ## Build Hooks for FastBoot
 
 ### Disabling incompatible dependencies
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "ember-cli-lodash-subset": "2.0.1",
     "ember-cli-preprocess-registry": "^3.1.0",
     "ember-cli-version-checker": "^2.1.0",
-    "fastboot": "github:rondale-sc/fastboot#utilize-rehydration-serialization-from-glimmer",
+    "fastboot": "^1.1.4-beta.1",
     "fastboot-express-middleware": "^1.1.0",
     "fastboot-transform": "^0.1.2",
     "fs-extra": "^4.0.2",
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "ember-cli-lodash-subset": "2.0.1",
     "ember-cli-preprocess-registry": "^3.1.0",
     "ember-cli-version-checker": "^2.1.0",
-    "fastboot": "^1.1.3",
+    "fastboot": "github:rondale-sc/fastboot#utilize-rehydration-serialization-from-glimmer",
     "fastboot-express-middleware": "^1.1.0",
     "fastboot-transform": "^0.1.2",
     "fs-extra": "^4.0.2",
--- a/vendor/experimental-render-mode-rehydrate.js
+++ b/vendor/experimental-render-mode-rehydrate.js
@@ -2,7 +2,11 @@
   if (typeof FastBoot === 'undefined') {
     var current = document.getElementById('fastboot-body-start');
 
-    if (current && Ember.ViewUtils.isSerializationFirstNode(current.nextSibling)) {
+    if (
+      current &&
+      typeof Ember.ViewUtils.isSerializationFirstNode === 'function' &&
+      Ember.ViewUtils.isSerializationFirstNode(current.nextSibling)
+    ) {
       Ember.ApplicationInstance.reopen({
         _bootSync: function(options) {
           if (options === undefined) {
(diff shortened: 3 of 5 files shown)


vendor/experimental-render-mode-rehydrate.js
```diff
@@ -0,0 +1,25 @@
+(function() {
+  if (typeof FastBoot === 'undefined') {
+    var current = document.getElementById('fastboot-body-start');
+
+    if (current && current.nextSibling.nodeValue === '%+b:0%') {
+      Ember.ApplicationInstance.reopen({
```
rwjblue: Is this how we plan to do this long term? Do we have a plan to remove this reopen?

package.json
```diff
@@ -29,7 +29,7 @@
     "ember-cli-lodash-subset": "2.0.1",
     "ember-cli-preprocess-registry": "^3.1.0",
     "ember-cli-version-checker": "^2.1.0",
-    "fastboot": "^1.1.3",
+    "fastboot": "github:rondale-sc/fastboot#utilize-rehydration-serialization-from-glimmer",
```
rwjblue: Just notating that this will need to be bumped to a release version before landing...

vendor/experimental-render-mode-rehydrate.js
```diff
@@ -0,0 +1,25 @@
+(function() {
+  if (typeof FastBoot === 'undefined') {
+    var current = document.getElementById('fastboot-body-start');
+
+    if (current && current.nextSibling.nodeValue === '%+b:0%') {
```
rwjblue: We _really really_ need a way to avoid encoding this, can you make an issue in glimmer-vm to expose a `isRehydrationComment` function that both Ember and ember-cli-fastboot can use? I'm concerned with the serialize markers changing over time (they have already changed once!?!) and then ember-cli-fastboot will have issues with cross-ember compat...

rondale-sc: I totally agree! I'll make the issue.

rondale-sc:
Closed and reopened to trigger another Travis build

rondale-sc:
@rwjblue This is ready once https://github.com/ember-fastboot/fastboot/pull/185 lands

rondale-sc:
@kratiahuja @rwjblue I'm not sure why these tests are failing.  They pass locally, and the only difference seems to be capitalization of the Content Type Header in the acceptance tests

vendor/experimental-render-mode-rehydrate.js
```diff
@@ -0,0 +1,25 @@
+(function() {
+  if (typeof FastBoot === 'undefined') {
+    var current = document.getElementById('fastboot-body-start');
+
+    if (current && Ember.ViewUtils.isSerializationFirstNode(current.nextSibling)) {
+      Ember.ApplicationInstance.reopen({
```
kratiahuja: Don't we need to turn off this: https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/app/instance-initializers/clear-double-boot.js ?

vendor/experimental-render-mode-rehydrate.js
```diff
@@ -0,0 +1,25 @@
+(function() {
+  if (typeof FastBoot === 'undefined') {
+    var current = document.getElementById('fastboot-body-start');
+
+    if (current && Ember.ViewUtils.isSerializationFirstNode(current.nextSibling)) {
```
kratiahuja: Why not provide a mixin for folks to add in their app.js itself?

vendor/experimental-render-mode-rehydrate.js
rondale-sc: The clear-double-boot checks to see if `fastboot-body-start` is present.  If we detect true with `Ember.ViewUtils.isSerializationFirstNode` we eventually delete `fastboot-body-start` which prevents the clear-double-render.

kratiahuja:
General question: How will apps opt in or opt out of rehydration? If there are issues, I would rather not have apps break and are forced to not use fastboot.

vendor/experimental-render-mode-rehydrate.js
rondale-sc: So long as `Ember.ViewUtils.isSerializationFirstNode` is present consumers of fastboot will never have to worry about this line of code unless fastboot actually encodes the DOM the browser version of Ember receives with the glimmer-vm's serialization element builder.  And if the the `isSerializationFirstNode` triggers true then we know the user has opted in.

rondale-sc:
@kratiahuja The way to opt-in once this PR is merged would be to ensure the `process.env.EXPERIMENTAL_RENDER_MODE_SERIALIZE` is set to `true`.  If that happens all the rest of the pipeline for rehydration should just work based off of that.

vendor/experimental-render-mode-rehydrate.js
kratiahuja: Is the `Ember.ViewUtils.isSerializationFirstNode` a public API?

kratiahuja:
We discussed this also needs doc updates :)

vendor/experimental-render-mode-rehydrate.js
rondale-sc: @kratiahuja It is public.

rwjblue: @rondale-sc - Can you tweak this to be safe when `Ember.ViewUtils.isSerializationFirstNode` is not present (e.g. you aren’t using >= some 3.2ish canary build ATM)

rondale-sc: @rwjblue Updated to ensure `typeof Ember.ViewUtils.isSerializationFirstNode === 'function'`

rwjblue: Thank you!

README.md
```diff
@@ -399,6 +399,41 @@ export default  Ember.Route.extend({
...
+
+The rehydration feature means that the Glimmer VM can take a DOM tree
+created using Server Side Rendering (SSR) and use it as the starting
+point for the append pass.
+
+See details here:
+
+https://github.com/glimmerjs/glimmer-vm/commit/316805b9175e01698120b9566ec51c88d075026a
+
+In order to utilize rehydration in Ember.js applications we need to ensure that
```
kratiahuja: Don't you need to mention from which version it is available?

kratiahuja:
`fastboot@1.1.4-beta.1` is released.

README.md
rondale-sc: Definitely!  I just added a small blurb about fastboot / ember.js requirements to address this concern.

README.md
```diff
@@ -399,6 +399,43 @@ export default  Ember.Route.extend({
...
+```
+
+This flag is read by Ember CLI Fastboot's dependency; fastboot to alert it to
+produce DOM with the glimmer-vm's serialization element builder.  This addon
+(Ember CLI Fastboot) then uses a utility function from glimmer-vm that allows
+it to know whether or not the DOM it received in the browser side was generated
+by the serialization builder.  If it was, it tells the Ember.js Application to
+use the rehydration builder and your application will be using rehydration.
+
+Rehydration is only compatible with fastboot > 1.1.4, and Ember.js > 3.2.
```
kratiahuja: This needs to be 1.1.4-beta.1

package.json
kratiahuja: Can you change this to `1.1.4-beta.1`?

bantic:
This needs an update to the yarn.lock also — I'll push that in a moment

bantic:
Pushed the updated yarn.lock. @kratiahuja This should be good to go now as soon as tests pass

kratiahuja:
thank you @bantic . I'll merge to master and cut a release for beta. I will also create a branch for stable if required and keep that seperate.

kratiahuja:
Thank you @rondale-sc and @bantic ! Merging this and releasing.

kratiahuja:
`ember-cli-fastboot@1.1.4-beta.1` is published.

PR #583, opened 2018-03-10
Fix tests

rondale-sc:
It is unknown right now as to why these are suddenly lowercase, this
directly reverts a change from about a year ago by Stef.

https://github.com/ember-fastboot/ember-cli-fastboot/commit/67102c964e6b276e3f318f01eca56944fe337e3c#diff-c2a8324c5e490c750018e7ed31eb42e9

Previously somewhere in our middleware stack upcased the `UTF-8` portion
of the Content Type header this doesn't seem to be happening anymore.

This fix seems innocuous for now.  Perhaps worth investigation later.

### DIFF
--- a/test/root-url-test.js
+++ b/test/root-url-test.js
@@ -63,7 +63,7 @@ describe('rootUrl acceptance', function() {
     })
       .then(function(response) {
         expect(response.statusCode).to.equal(200);
-        expect(response.headers["content-type"]).to.eq("text/html; charset=UTF-8");
+        expect(response.headers["content-type"]).to.eq("text/html; charset=utf-8");
         expect(response.body).to.contain("<!-- EMBER_CLI_FASTBOOT_BODY -->");
       });
   });
--- a/test/serve-assets-test.js
+++ b/test/serve-assets-test.js
@@ -30,7 +30,7 @@ describe('serve assets acceptance', function() {
     return request('http://localhost:49741/assets/vendor.js')
       .then(function(response) {
         expect(response.statusCode).to.equal(200);
-        expect(response.headers["content-type"]).to.eq("application/javascript; charset=UTF-8");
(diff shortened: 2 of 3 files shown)


kratiahuja:
Thanks @rondale-sc ! was there a bump in express server or something that is causing this?

rondale-sc:
@bantic did some digging into the http spec to determine which is actually valid it mentions explicitly that Content Type charset parameter value is case insensitive.  See here for more information:

https://tools.ietf.org/html/rfc2616#section-3.4

rondale-sc:
@kratiahuja Unsure as to whether the the express version bumped it, still kinda digging into this a little bit.  Will post back with any findings.

bantic:
After looking carefully through the deps I still haven't found a smoking gun where the charset case changed from "UTF-8" to "utf-8". 
Did a little spelunking through the http spec and finally came across the [IANA Charset Doc](https://www.iana.org/assignments/character-sets/character-sets.xhtml) which states:

> no distinction is made between use of upper and lower case letters

So I think the best thing to do here is to change the tests to compare down-cased content-type to "...; charset=utf-8"

I'll add a quick follow-on commit to this PR to make that change.

FWIW, tests within express itself don't seem to agree on the casing of "utf-8" (here's a [test with upper-case](https://github.com/expressjs/express/blob/c0136d8b48dd3526c58b2ad8666fb4b12b55116c/test/res.download.js#L18), and here's [one with lower-case](https://github.com/expressjs/express/blob/40e04ec7a6d365a7e083b0fdf7f9d2c7afc036a0/test/res.format.js#L168)).

kratiahuja:
I think we probably need to make this check case insensitive. I will go ahead and merge this and work on making the tests more robust.

bantic:
@kratiahuja I've got a PR coming in a few minutes that changes the tests to do case-insensitive matching for the content-type header

kiwiupover:
Thank for landing this pr.

PR #584, opened 2018-03-10
Use case-insensitive matching when comparing content-type

bantic:
See #583

The "charset=utf-8" in content-type sometimes changes in tests depending
on what express middleware serves the response.

The charset value is case-insensitive according to IANA (see:
https://www.iana.org/assignments/character-sets/character-sets.xhtml)

This adds chai-string (http://chaijs.com/plugins/chai-string/) as a devDep and
uses its `equalIgnoreCase` matcher in tests that check the value of
the content-type header.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "broccoli-test-helper": "^1.1.0",
     "chai": "^4.1.0",
     "chai-fs": "^2.0.0",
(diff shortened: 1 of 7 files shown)


rondale-sc:
@bantic I think this looks pretty great, thanks for doing the research on this! :)

PR #585, opened 2018-03-10
Support module unification app in `/src/main`

iezer:
@rwjblue  @kratiahuja FYI this change is need in order to use fastboot with a module unification app.

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,7 +92,8 @@ module.exports = {
     }
 
     if (type === 'app-boot') {
-      return fastbootAppModule(config.modulePrefix, JSON.stringify(config.APP || {}));
+      const isModuleUnification = (typeof this.project.isModuleUnification === 'function') && this.project.isModuleUnification();
+      return fastbootAppModule(config.modulePrefix, JSON.stringify(config.APP || {}), isModuleUnification);
     }
 
     // if the fastboot addon is installed, we overwrite the config-module so that the config can be read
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -77,6 +77,17 @@ module.exports = {
     }
   },
 
+  /**
+     Returns whether this app is using a module unification format.
+     @private
+     @method isModuleUnification
+     @return {Boolean} Whether this app is using a module unification format.
+  */
+  isAppModuleUnification() {
```
rwjblue: Should be using `this.project.isModuleUnificatiob()` (no need for a new helper method AFAICT)

iezer: @rwjblue Good call, I've made that change.,

index.js
```diff
@@ -92,7 +92,7 @@ module.exports = {
     }
 
     if (type === 'app-boot') {
-      return fastbootAppModule(config.modulePrefix, JSON.stringify(config.APP || {}));
+      return fastbootAppModule(config.modulePrefix, JSON.stringify(config.APP || {}), this.project.isModuleUnification());
```
rwjblue: Hmm, we may need to handle the case where the function doesn’t exist (IIRC it only exists on ember-cli master).

iezer: as in `this.project.isModuleUnification && this.project.isModuleUnification()` ?

iezer: or perhaps `(typeof this.project.isModuleUnification === 'function') && this.project.isModuleUnification()`

iezer: fixed.

kratiahuja:
How soon does need to land?

rwjblue:
@kratiahuja - yeah, agreed

kratiahuja:
Ok I will release this first and then release beta.

PR #604, opened 2018-08-08
Use this to prevent memory leak of application instance

bobisjan:
screen shot 2018-08-08 at 17 50 48

### DIFF
--- a/addon/instance-initializers/clear-double-boot.js
+++ b/addon/instance-initializers/clear-double-boot.js
@@ -29,7 +29,7 @@ export default {
 
       instance.didCreateRootView = function() {
         clearHtml();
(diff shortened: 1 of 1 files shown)


addon/instance-initializers/clear-double-boot.js
```diff
@@ -29,7 +29,7 @@ export default {
 
       instance.didCreateRootView = function() {
         clearHtml();
-        originalDidCreateRootView.apply(instance, arguments);
+        originalDidCreateRootView.apply(this, arguments);
```
snewcomer: @bobisjan Do you still have this issue?  Is this due to babel transpilation?  i.e. what happens if we convert the function to arrow syntax so it operates in the context of the enclosing scope?

PR #636, opened 2018-10-30
Skip shoebox elements when clearing pre-rendered elements

paulgoetze:
Addresses the shoebox issue described in https://github.com/ember-fastboot/ember-cli-fastboot/issues/565#issuecomment-430364978.

PR #654, opened 2019-01-28
Fix for #636

### LINKED ISSUE #653: #636 (version 2.0.1) broke IE11 support due to Array.from
#636 (version 2.0.1) introduced the check:

`!Array.from(shoeboxNodes).includes(nextNode)`

IE11 doesn't support `Array.from`. My understanding is that Ember and Fastboot officially support IE11, so this is a regression.

PR with a fix coming...

samcic:
Replaced `Array.from` usage to restore IE11 support. Fixes #653


rwjblue:
ember-cli-fastboot 2.0.4 published 🎉

PR #662, opened 2019-02-20
Allow using the dummy app for testing purposes.

nathanhammond:
This is a straw man in support of the argument I make here: https://github.com/ember-fastboot/ember-cli-fastboot/pull/467#issuecomment-465369542

I posit that being able to use `ember-cli-fastboot` as a dependency of an addon solely for testing is beneficial. I've scoped this change to allow for exactly that and _no more_, in keeping with the decision in #467.

### DIFF
--- a/index.js
+++ b/index.js
@@ -29,7 +29,21 @@ function getVersionChecker(context) {
   return checker;
 }
 
+function getClosestHost(node) {
+  const projectReference = node.parent && node.parent.parent;
 
+  if (
+    projectReference &&
+    !node.app &&
+    !projectReference.parent &&
+    node.parent.app &&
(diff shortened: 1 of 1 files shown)


nathanhammond:
This goes wrong if you've got `ember-cli-fastboot` as a `dependency` of the addon, not a `devDependency`. If you mark it as a `devDependency` it will not be discovered by `ember-cli` in the addon pass, and work correctly. If you mark it as a `dependency` it will do the "wrong" thing of making `ember-cli-fastboot` a dependency of the addon, which is attached to the app project.

This is because even in addons the dummy app *is* the project. `devDependencies` are treated as top-level app (project) dependencies while `dependencies` are discovered from the addon itself.

PR #671, opened 2019-03-13
Sketch of lockfile generation.

### LINKED ISSUE #668: Generate lockfile at build time.
When creating a production build of an application, the build should be repeatable. In order for a build to be repeatable, any dependencies and transitives should be locked.

This is not presently possible using `ember-cli-fastboot` as it only spits out a `package.json` file. We should extend it to also emit a `package-lock.json` or `yarn.lock`.

nathanhammond:
This is a sketch of what lockfile generation could look like at build time.

Fixes #668


rwjblue:
This looks reasonable, but I don't think we should do this on every build/rebuild. Instead it should only be for production builds.

PR #674, opened 2019-03-20
Updating testing matrix to drop node 6 and add node 10 & 11

kiwiupover:
Updated test-metrix:
- Drop node 6
- Add node 10
- Add node 11
- Update NPM to 6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 ---
 language: node_js
 node_js:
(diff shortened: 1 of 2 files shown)


rwjblue:
FWIW, I don't think we can drop Node 6 until its actually EOL'ed.

kiwiupover:
Fair enough

PR #679, opened 2019-03-29
Get tests passing again by testing against ember-data@3.8.0 instead of master

kiwiupover:
https://github.com/tomdale/ember-cli-addon-tests/pull/208

ember-cli-addon-test by default tests against ember-data#master this PR removes #master and uses ^3.8.0

adds node 10 to the test metrics
uses npm 6 or greater


rwjblue:
Pushed a commit to use the newly released version of ember-cli-addon-tests (0.11.1).

PR #695, opened 2019-05-07
Drop Node 6, 8, 9, 11, and 13 support.

rwjblue:


package.json
```diff
@@ -74,7 +74,7 @@
     "rsvp": "^4.8.3"
   },
   "engines": {
-    "node": "6.* || 8.* || >= 10.*"
+    "node": "8.* || 10.* || >= 12.*"
```
stefanpenner: This also drops support for node 11, which doesn't EOL until 2019-06-01. Was that the intention?

rwjblue: Yeah, but I can roll it back

rwjblue: Updated to bring back Node 11 support

SergeAstapov: 2019-06-01 has passed, shall Node.js 11 support be dropped?

appveyor.yml
```diff
@@ -4,8 +4,8 @@
 environment:
   MOCHA_REPORTER: "mocha-appveyor-reporter"
   matrix:
-    - nodejs_version: "6"
-    - nodejs_version: "8"
+    - nodejs_version: "10"
```
scalvert: Shouldn't this include node 8, as per the .travis.yml?

PR #699, opened 2019-05-29
Terrible, Awful, No Good, Very Bad Markup Crashes App

noslouch:
When rendering dynamic markup pulled from a database, we don't have a guarantee things will end up in the DOM where we want them to be.

Ember enforces good markup during the build, but there's no way to know if the following template will have valid markup:
```handlebars
// e.g. app/templates/article.hbs
<main id="content">
  {{{model.body}}}
</main>
```

The browser is a tolerant place, and it will do whatever it needs to do to create a valid DOM tree, even if that means swallowing up elements. A certain type of bad markup can wreak havoc on desired trees.

For example, let's say your fastboot returns the following HTML document to your browser:
```html
<!-- snip... -->
<body>
  <script type="x/boundary" id="fastboot-body-start"></script>
  <div>
    <p>
      Great article content here.
    </p>
    <em>With additional reporting<em> <---- ¯\_(ツ)_/¯
  </div>
  <script type="x/boundary" id="fastboot-body-end"></script>
</body>
```

When a browser parses that (at least when Chrome does), here's the DOM tree you wind up with:
```html
<body>
  <script type="x/boundary" id="fastboot-body-start"></script>
  <div>
    <p>
      Great article content here.
    </p>
    <em>
      With additional reporting
      <em></em>
    </em>
  </div>
  <em>
    <em>
      <script type="x/boundary" id="fastboot-body-end"></script>
    </em>
  </em>
</body>
```
Yikes.

When the `clear-double-boot` initializer runs, it assumes that the start and end markers have the same parent, which is sometimes not the case, as seen above.

Does this proposed solution work for you all? I know it might open a can of worms as far as working around DOM quirks, but this seems like a particularly nasty case that we can fix to ensure that the app finishes booting.

### DIFF
--- a/addon/instance-initializers/clear-double-boot.js
+++ b/addon/instance-initializers/clear-double-boot.js
@@ -22,7 +22,7 @@ export function clearHtml() {
       parent.removeChild(current);
       current = nextNode;
     } while (nextNode && nextNode !== endMarker && shoeboxNodesArray.indexOf(nextNode) < 0);
-    parent.removeChild(endMarker);
+    endMarker.parentElement.removeChild(endMarker);
   }
 }
 export default {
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you @noslouch!

Issue #700, opened 2019-06-03
FastBoot + Embroider issues


dnalagatla:
This is a meta issue to track tickets related to FastBoot + Embroider build.

Issues: 

- [ ] [Embroider + Fastboot: `dist` folder doesn't include <app-name>.js file](https://github.com/embroider-build/embroider/issues/184)

- [ ] [Embroider + Fastboot: Build not invoking `app-boot` in contentFor hook.](https://github.com/embroider-build/embroider/issues/183)

PR #701, opened 2019-06-05
Added support to update manifest appfiles from index.html

dnalagatla:
In order to support Embroider buld: 

* Added postBuild hook to update `assetFiles` property in the fastboot manifest to load in the sandbox. 

~~* Also, as autoRun is false in fastboot and `app-boot` type not supported in `contentFor` hook.  Moved the logic to postBuild. Used postBuild hook for appending fastboot module and a check to create app when not in fastboot environment.~~ - Merged as part of this pr - https://github.com/ember-fastboot/ember-cli-fastboot/pull/702

#700 - https://github.com/embroider-build/embroider/issues/184

Tested Application using embroider build and application not using embroider build.

### DIFF
--- a/index.js
+++ b/index.js
@@ -346,6 +346,7 @@ module.exports = {
   },
 
   postBuild(result) {
+    this._updateAppFilesInManifest(result.directory);
     if (this.fastboot) {
       // should we reload fastboot if there are only css changes? Seems it maynot be needed.
       // TODO(future): we can do a smarter reload here by running fs-tree-diff on files loaded
@@ -376,5 +377,51 @@ module.exports = {
   
   _isModuleUnification() {
     return (typeof this.project.isModuleUnification === 'function') && this.project.isModuleUnification();
+  },
+
+  _updateAppFilesInManifest(appDir) {
+    const pkgPath = path.join(appDir, 'package.json');
+    const pkg = require(pkgPath);
+    const {
+      fastboot
+    } = pkg;
+
+    if(fastboot) {
+      const { manifest: { appFiles: manifestAppFiles } } = fastboot;
+      let appFilePath = '';
+      if(Array.isArray(manifestAppFiles) && typeof manifestAppFiles[0] === 'string') {
+        appFilePath = path.resolve(appDir, manifestAppFiles[0]);
+      }
+      // If app file in package.json doesn't exist, update manifest.appFiles in package.json
+      // with appFiles defined in index.html. This check will ensure the existing behavior is intact and
+      // only update from index.html for embroider flow.
+      if(appFilePath && !existsSync(appFilePath)) {
+        const appFiles = this._getAppFilesFromIndexHtml(appDir);
+        manifestAppFiles.splice(0, 1, ...appFiles);
+        fs.writeFileSync(pkgPath, JSON.stringify(pkg));
+      }
+    }
+  },
+
+  _getAppFilesFromIndexHtml(appDir) {
+    const cheerio = require('cheerio');
+    const indexHtml = fs.readFileSync(path.join(appDir, 'index.html'));
+    const $ = cheerio.load(indexHtml);
+    const scriptFileNameRegEx = /([a-zA-Z0-9_\.\-\(\):])+(\.js)/ig;
+    const filesToSkipRgex = /^vendor|^vendor-static|^ember-cli-live-reload/gi;
+    const appFiles = [];
+
+    $('script').each(function(i, elem) {
+      const src = $(elem).attr('src');
+      if (src) {
+        const fileName = src.match(scriptFileNameRegEx)[0];
+        const filePath = path.join(appDir, 'assets', fileName);
+        if (fileName && existsSync(filePath) && !filesToSkipRgex.test(fileName)) {
+          appFiles.push(path.relative(appDir, filePath));
+        }
+      }
+    });
+
+    return appFiles;
   }
 };
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "chai": "^4.1.2",
     "chai-fs": "^2.0.0",
     "chai-string": "^1.4.0",
+    "cheerio": "^1.0.0-rc.3",
     "co": "^4.6.0",
     "ember-ajax": "^3.1.0",
     "ember-cli": "~3.3.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1635,6 +1635,10 @@ body@^5.1.0:
     raw-body "~1.1.0"
     safe-json-parse "~1.0.1"
 
+boolbase@~1.0.0:
+  version "1.0.0"
+  resolved "https://registry.yarnpkg.com/boolbase/-/boolbase-1.0.0.tgz#68dff5fbe60c51eb37725ea9e3ed310dcc1e776e"
+
 bower-config@^1.3.0:
   version "1.4.1"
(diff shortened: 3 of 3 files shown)


dnalagatla:
@rwjblue @stefanpenner

index.js
```diff
@@ -209,6 +205,7 @@ module.exports = {
   },
 
   treeForPublic(tree) {
+    debugger;
```
tomdale: 👀

tomdale:
Cheerio is a pretty large dependency for FastBoot to take on.

Want to get feedback from @stefanpenner @ef4 @rwjblue that the correct architecture is for us to parse the HTML file (vs. emitting some kind of JSON manifest), and that cheerio is the simplest way to do that.

index.js
dnalagatla: removed that

rwjblue:
We should be able to parse with `simple-html-tokenizer`, which is already in our dep graph...

ef4:
Yes, parsing the HTML file is the way to go. It aligns fastboot with web standards instead of our own ember-only standard. It means that the output from Parcel or Webpack can go directly to fastboot and fastboot will be able to know how to boot it.

(The other piece of this story is that any browser-only or fastboot-only code will be guarded by the macro system. In development we can do a single build that keeps both branches in, while running correctly in both environments. In prod we can produce both optimized builds.)

dnalagatla:
> Cheerio is a pretty large dependency for FastBoot to take on.
> 
> Want to get feedback from @stefanpenner @ef4 @rwjblue that the correct architecture is for us to parse the HTML file (vs. emitting some kind of JSON manifest), and that cheerio is the simplest way to do that.

I am open to using any other library here for parsing the HTML. As this a build time parsing and cheerio being the dev dependency I don't see this causing issue during runtime.

One drawback of parsing HTML would be related to production builds replacing asset URLs with hashed CDN URLs. In that case, this postBuild processing would fail as hashed assets wouldn't be present in app's package on the disk. Any thoughts about this scenario @tomdale @stefanpenner @ef4 @rwjblue

rwjblue:
> One drawback of parsing HTML would be related to production builds replacing asset URLs with hashed CDN URLs. 

@ef4 - Have you considered this case? Is this something that we would have to track and _reverse_ the work of the fingerprinter?

ef4:
CDN prefixing and fingerprinting are two different concerns.

Fingerprinting does not need to be reversed (and can't be, in the case of embroider, because it's up to the stage3 packager to do whatever it wants with filenames).

The way we do CDN prefixing today, yes, it would need to be reversed. But that is much easier (it only requires one argument to fastboot, the prefix it's allowed to strip off URLs to make them local again). Arguably we don't even need that, because fastboot can see all the assets on disk and can correlate URL suffixes with filenames, allowing it to discover any CDN prefixes on the fly.

rwjblue:
Now that #702 is landed, can you rebase?

dnalagatla:
> Now that #702 is landed, can you rebase?

Thank you @rwjblue I did the rebase.

index.js
```diff
@@ -327,8 +328,7 @@ module.exports = {
   },
 
   postBuild(result) {
-    let appFilePath = stripLeadingSlash(this.app.options.outputPaths.app.js);
-    this._appendAppBoot(result.directory, appFilePath);
+    this._updateAppFilesInManifest(result.directory);
```
rwjblue: Is there a way for us to only do this in Embroider land?

index.js
```diff
@@ -19,6 +19,7 @@ const Funnel = require('broccoli-funnel');
 const p = require('ember-cli-preprocess-registry/preprocessors');
 const fastbootTransform = require('fastboot-transform');
 const existsSync = fs.existsSync;
+const cheerio = require('cheerio');
```
rwjblue: We should swap to using simple-html-tokenizer to parse, I'll try to whip up a demo to show what I mean.

index.js
ef4: There's no obvious way and I don't really want to give one to people. Running Embroider and a classic build within the same broccoli system is a valid thing to want to do.

dnalagatla: @ef4 any reasons for not to provide a way to check the embroider land. It will make it easier for Fastboot's to switch between classic build flow and changes needed to be made for embroider. Also, will help in isolating the build issue specific to embroider.

ef4: There's really no such thing as "outside" vs "inside" embroider. The addons all get instantiated *before* any embroider code even runs, and that is intentional. Also, the last time we exposed a global flag like this inside broccoli (the one for checking for fastboot) we regretted it, because it's just not really how broccoli works.

So no, I really don't want v1-formatted addons to start adding code to behave special under embroider. The only supported way to do that is to ship as a native V2 addon, after we merge [RFC 507](https://github.com/emberjs/rfcs/pull/507).

Also, this change is an improvement for *all* builds, not just embroider. It moves us in a better architectural direction.

dnalagatla: Thank you @ef4. This makes sense.

dnalagatla:
Hi @kratiahuja @tomdale @rwjblue  @ef4, There is a test failure related to this test - https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/test/package-json-test.js#L232. This test to custom output paths configured in `ember-cli-build.js`

Here is the configuration of custom output path used for above test - https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/test/fixtures/customized-outputpaths/ember-cli-build.js#L5

Not sure how this scenario be handled in terms of index.html,  should we expect the custom paths to be present in index.html. As I am not familiar with ember-cli set up for that. please let me know your thoughts.

rwjblue:
Chatted with @dnalagatla and @mansona in today's FastBoot meeting, in order to work around that failing test we will need to keep the existing behaviors around as a fallback.

dnalagatla:
> Chatted with @dnalagatla and @mansona in today's FastBoot meeting, in order to work around that failing test we will need to keep the existing behaviors around as a fallback.

@rwjblue I made updates to keep the current low intact and only get the app files from index.html when <app>.js file doesn't exist. Test related custom output paths is passing now.

dnalagatla:
Rebased with master and resolved conflicts. @rwjblue please review when you get a chance

index.js
```diff
@@ -376,5 +377,48 @@ module.exports = {
...
+  _updateAppFilesInManifest(appDir) {
+    const pkgPath = path.join(appDir, 'package.json');
+    const pkg = require(pkgPath);
+    const {
+      fastboot
+    } = pkg;
+
+    if(fastboot) {
+      const { manifest: { appFiles: manifestAppFiles } } = fastboot;
+      const appFilePath = path.resolve(appDir, manifestAppFiles[0]);
```
rwjblue: Does this error if `manifestAppFiles` is an empty array? Would it be empty in the embroider case?

dnalagatla: Thank you @rwjblue.  That is correct. if `manifestAppFiles` is empty, the `resolve` method will throw an error. But In the first stage of the build of embroider we will always expect generated app paths to be present.

But,  I added a defensive check that if `manifestAppFiles` is an Array and the first element is a string to make sure there is no error thrown here.

package.json
```diff
@@ -45,6 +45,7 @@
     "chai": "^4.1.2",
     "chai-fs": "^2.0.0",
     "chai-string": "^1.4.0",
+    "cheerio": "^1.0.0-rc.3",
```
stefanpenner: I believe this should be a dependency not a devDependency

stefanpenner: https://github.com/ember-fastboot/ember-cli-fastboot/pull/725

stefanpenner:
This code does not appear to support rebuilds, when a rebuild occurs it appears to go back to not reading the `index.html`

Issue #706, opened 2019-07-13
Is the #master branch broken?


kiwiupover:
@dnalagatla can you confirm that the master branch is building the ember app correctly? 

## Failing from `#master` 
https://github.com/kiwiupover/fastboot-test/pull/1
```json
"ember-cli-fastboot": "github:ember-fastboot/ember-cli-fastboot#master",
```
https://github.com/kiwiupover/fastboot-test/pull/1/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R30

This does the SSR step correctly but the ember app is not booted. Try opening the ember inspector. 

## Working from `v2.1.1`

https://github.com/kiwiupover/fastboot-test
```json
"ember-cli-fastboot": "^2.1.1",
```

dnalagatla:
@kiwiupover  I think the issue related to my last PR where I move the app-boot code to `postBuild` hook. 

I created a new PR for the fix - https://github.com/ember-fastboot/ember-cli-fastboot/pull/707

I tested my change with fastboot-test and it worked. Thank you reporting the issue.

joshkg:
I'm seeing this issue too in 2.1.2 - I can see a new app boot file with the app name is generated during production build that contains the JS to start the app, but that script link is not appended to the index.html, and thus the ember app does not boot.

dnalagatla:
Hi @kratiahuja, I have a PR to revert embroider related changes to fix the issue related to productions builds - https://github.com/ember-fastboot/ember-cli-fastboot/pull/713

Please help with merging this PR when you get a chance. Thank you

dnalagatla:
> I'm seeing this issue too in 2.1.2 - I can see a new app boot file with the app name is generated during production build that contains the JS to start the app, but that script link is not appended to the index.html, and thus the ember app does not boot.

Thank you for reporting. I have created a PR to revert the changes related to embroider build related support. - https://github.com/ember-fastboot/ember-cli-fastboot/pull/713

dnalagatla:
Thank you @kratiahuja for merging the changes. The Issue should be fixed in v2.1.3

PR #707, opened 2019-07-13
Checking if runningTests not defined or set to false

dnalagatla:
In embroider build `runningTests` is not defined whereas in ember-cli  build it is set to false. #706


dnalagatla:
cc: @rwjblue @kiwiupover

PR #714, opened 2019-07-28
Embroider Support: Moved app factory module out of fastboot-app-module

dnalagatla:
* Cleaned up fastboot-app-module.
* Added Fastboot-app-factory-plugin to insert app factory module as part of <app>-fastboot.js
* This clean up ensures fastboot specific App-factory code is not shipped as part of browser assets.
* With Embroider not supporting `app-boot` type contentFor, this change ensures the insertion of the app-factory module is not dependent on `app-boot` step in the build process.

WIP: #700

### DIFF
--- a/index.js
+++ b/index.js
@@ -9,8 +9,9 @@ const FastBootExpressMiddleware = require('fastboot-express-middleware');
 const FastBoot = require('fastboot');
 const chalk = require('chalk');
 
-const fastbootAppModule = require('./lib/utilities/fastboot-app-module');
+const fastbootAppBoot = require('./lib/utilities/fastboot-app-boot');
 const FastBootConfig = require('./lib/broccoli/fastboot-config');
+const fastbootAppFactoryModule = require('./lib/utilities/fastboot-app-factory-module');
 const migrateInitializers = require('./lib/build-utilities/migrate-initializers');
 const SilentError = require('silent-error');
 
@@ -107,8 +108,8 @@ module.exports = {
     }
 
     if (type === 'app-boot') {
-      const isModuleUnification = (typeof this.project.isModuleUnification === 'function') && this.project.isModuleUnification();
-      return fastbootAppModule(config.modulePrefix, JSON.stringify(config.APP || {}), isModuleUnification);
+      const isModuleUnification = this._isModuleUnification();
+      return fastbootAppBoot(config.modulePrefix, JSON.stringify(config.APP || {}), isModuleUnification);
     }
 
     // if the fastboot addon is installed, we overwrite the config-module so that the config can be read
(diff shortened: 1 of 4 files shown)


dnalagatla:
@rwjblue @stefanpenner  Created this PR to clean up fastboot-app-module code that gets inserted as part of app.js. 

I made changes to move the app-factory to a separate module and created a broccoli plugin to generate the content and merged as part of `app-fastboot.js`.

index.js
```diff
@@ -174,10 +175,11 @@ module.exports = {
    */
   _getFastbootTree() {
     const appName = this._name;
+    const isModuleUnification = (typeof this.project.isModuleUnification === 'function') && this.project.isModuleUnification();
```
SergeAstapov: just out of curiosity, does it make sense to put new code for module unification [considering it's status](https://blog.emberjs.com/2019/03/11/update-on-module-unification-and-octane.html)?

index.js
```diff
@@ -9,8 +9,9 @@ const FastBootExpressMiddleware = require('fastboot-express-middleware');
 const FastBoot = require('fastboot');
 const chalk = require('chalk');
 
-const fastbootAppModule = require('./lib/utilities/fastboot-app-module');
+const fastbootAppBoot = require('./lib/utilities/fastboot-app-boot');
 const FastBootConfig = require('./lib/broccoli/fastboot-config');
+const FastBootAppFactory = require('./lib/broccoli/fastboot-app-factory-plugin');
```
SergeAstapov: out of curiosity, what is the benefit to have such broccoli plugin over [broccoli-file-creator](https://github.com/rwjblue/broccoli-file-creator)?

index.js
dnalagatla: > just out of curiosity, does it make sense to put new code for module unification [considering it's status](https://blog.emberjs.com/2019/03/11/update-on-module-unification-and-octane.html)?

Hi, This was not a new code but exiting check used in contentFor hook. I copy-pasted that here. To make it cleaner, I moved the common code of isModuleUnification to a separate method. The Goal of this PR was not change exiting behavior but clean up the code inserted as part of `app-boot` hook.

We should consider removing the check for module unification as a sperate work item.

index.js
dnalagatla: > out of curiosity, would it make sense/be possible to use [broccoli-file-creator](https://github.com/rwjblue/broccoli-file-creator) over new broccoli plugin?

Thank you for your suggestion. I wasn't aware of that broccoli plugin. I gave a glance and looks like I can use this plugin instead of creating a new one. I will give a try.

index.js
SergeAstapov: Totally makes sense, thank you for the note @dnalagatla!

index.js
dnalagatla: I made changes to use ` broccoli-file-creator` plugin instead of creating a new plugin.

PR #718, opened 2019-08-09
Point to ember-data-storefront instead of ember-cached-shoe

danwenzel:
See https://github.com/Appchance/ember-cached-shoe/issues/11

As I stated there, ember-cached-shoe doesn't seem to be actively maintained anymore, and ember-data-storefront provides the same functionality out of the box.

So this PR changes the "Think out of the Shoebox" section to suggest ember-data-storefront instead of ember-cached-shoe.


rwjblue:
Thank you!

PR #727, opened 2019-09-12
[WIP] embroider work does not currently support rebuilds, this …

stefanpenner:
- [x] fix rebuild
- [x] add tests
- [x] drop cheerio in-favor of simple-html-tokenizer
- [x] confirm fixes https://github.com/embroider-build/embroider/pull/322
- [ ] work through the inline todos


lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const fs = require('fs');
+const findScriptSrcs = require('find-scripts-srcs-in-document');
+// TODO: review this
+const FILES_TO_SKIP = /vendor|vendor-static/gi;
```
stefanpenner: look at this

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,37 @@
...
+
+module.exports = function updateManifestFromHtml(root) {
+  const pkg = JSON.parse(fs.readFileSync(root + '/package.json', 'UTF8'));
+
+  // bail if not embroider (let's chat if this is appropriate);
+  if (!('version' in pkg['ember-addon'])) {
```
stefanpenner: @ef4 / @rwjblue thoughts on a more explicit check?

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,37 @@
...
+    return;
+  }
+
+  // otherwise, we must parse the index.html file to figure out what's next
+  const indexHtml = fs.readFileSync(root + '/index.html', 'UTF8')
+  // TODO: fix simple-html-tokenizer: https://github.com/tildeio/simple-html-tokenizer/pulls
```
stefanpenner: https://github.com/tildeio/simple-html-tokenizer/pull/71

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,38 @@
...
+  const pkg = JSON.parse(fs.readFileSync(root + '/package.json', 'UTF8'));
+
+  // bail if not embroider (let's chat if this is appropriate);
+  if (!('version' in pkg['ember-addon'])) {
+    return;
+  }
+
+  // otherwise, we must parse the index.html file to figure out what's next
+  const indexHtml = fs.readFileSync(root + '/index.html', 'UTF8')
+  // TODO: fix simple-html-tokenizer: https://github.com/tildeio/simple-html-tokenizer/pull/71
```
stefanpenner: fix simple-html-tokenizer: https://github.com/tildeio/simple-html-tokenizer/pull/71

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,38 @@
...
+    .replace('<!DOCTYPE html>','')
+
+  const appFiles = findScriptSrcs(indexHtml, findScriptSrcs.ignoreWithAttribute('data-embroider-ignore'))
+    // TODO: these FILES_TO_SKIP things seem over zealous
+    .filter( src => !FILES_TO_SKIP.test(src))
```
stefanpenner: feels overzealous

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,38 @@
...
+    // TODO: handle prefixed / custom origin paths / or files
+    .filter(src => fs.existsSync(`${root}/${src}`));
```
stefanpenner: handle prefixed/custom origin path  etc.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,38 @@
...
+
+  // TODO: this feels super janky, we need to figure out a better approach
+  const fastbootFile = pkg.fastboot.manifest.appFiles.find(x => /-fastboot.js$/.test(x))
```
stefanpenner: this feels risky, we should find a better approach

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+const fs = require('fs');
+const findScriptSrcs = require('find-scripts-srcs-in-document');
```
stefanpenner: tested lib that uses simply-html-tokenizer

test/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,125 @@
+'use strict';
+const { expect } = require('chai');
+const Project = require('fixturify-project');
+const fs = require('fs');
+const updateManifestFromHtml = require('../../lib/embroider/update-manifest-from-html')
+
+describe('embroider/update-manifest-from-html', function() {
```
stefanpenner: TODO: more tests

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,39 @@
...
+
+// TODO: hard coding these seem unfortunate, especially vendor-static which is non-standard
+const FILES_TO_SKIP = /vendor|vendor-static/gi;
+
+module.exports = function updateManifestFromHtml(root) {
+  const pkg = JSON.parse(fs.readFileSync(root + '/package.json', 'UTF8'));
+  const addon = pkg['ember-addon'];
+
+  // bail if not embroider (let's chat if this is appropriate);
+  if (typeof addon !== 'object' || !('version' in addon)) {
```
dnalagatla: I am also, curious to know if there is a better option for the check regarding embroider or classic.

ef4: Addons doing different things under embroider is a bad idea. 

The whole point of `@embroider/compat` is to support addons that don't know about embroider. If an addon wants to actually support embroider, it needs to ship as a v2 addon, thus avoiding `@embroider/compat` entirely and only using supported APIs in `@embroider/core`.

rwjblue: > If an addon wants to actually support embroider, it needs to ship as a v2 addon

Which is actually impossible since it isn't shipped yet 😉

dnalagatla: > . If an addon wants to actually support embroider, it needs to ship as a v2 addon, thus avoiding `@embroider/compat` entirely and only using supported APIs in `@embroider/core`

Could be a dumb question, but just curious -  what can be done to use V2 but still be backward compatible. I think Updating ember-cli-fastboot AddOn to use V2 would need  Major version bump.  And have to maintain both versions. I don't think that is scalable. 

@rwjblue please correct me if I am wrong.

dnalagatla: I am very tempted to use `process.env.EMBROIDER` and update the documentation to indicate, if we have to use embroider with fastboot then run `EMBROIDER=true ember s`.  And use that property to differentiate embroider and classic flow.

I am not happy with this approach but will cover our requirement and since embroider is an experimental build I think this will be ok.

dnalagatla: any thoughts?

dnalagatla:
This PR can be closed as all the necessary changes for fixing embroider flow is in this PR - https://github.com/ember-fastboot/ember-cli-fastboot/pull/729

lib/embroider/update-manifest-from-html.js
dnalagatla: I updated the PR - https://github.com/ember-fastboot/ember-cli-fastboot/pull/729, with changes related to using `process.env.EMBROIDER` property to check if this is embroider flow or not.

ef4: @dnalagatla it won't require a major bump, v2 addons can also be valid v1 addons because ember-cli respects the `ember-addon.main` key in package.json, which lets us distinguish the v1 entrypoint from the v2 build entrypoint (`ember-addon.build`) and the v2 runtime entrpoint (`main`).

@rwjblue there's nobody better positioned than the people in this thread to make v2 addons become fully implemented. There's not a lot of work required, since the whole system is already designed around their needs. But that is all beside my point: these changes aren't breaking as far as I know, if they're breaking show me how and we can adjust until they're not breaking. If you just want a feature flag inside ember-cli-fastboot to manage the rollout, fine, but don't name it "EMBROIDER" because that's just confusing. It's an ember-cli-fastboot specific feature flag.

dnalagatla: @ef4 There is a difference between the classic build and embroider build flow, as a result I was interested in having that flag: 

In Classic build flow the following are the steps occurs in ember-cli-fastboot build -
1. Ember-cli gives a flag to turn off autoBoot by setting this property `app.options.autoRun` and is set to false by default in include hook in `ember-cli-fastboot` - https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L73
2. Then in `contentFor` hook `app-boot` type is used to define custom app-boot logic that can support Browser and Fastboot (using appFactory). - https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L110

Embroider build doesn't support `app-boot` hook in `contentFor` to include custom app-boot logic needed for ember-cli-fastboot. - https://github.com/embroider-build/embroider/issues/183

As `app-boot`, we introduced a flag to disable-auto-boot in fastboot, but for that flag to be effective we would have to set `app.options.autoRun` to `true` which is only compatible with embroider flow but not with classic build flow of ember-cli-fastboot.

dnalagatla: Hi @ef4, are there any plans to support `app-boot` in embroider.

ef4:
Yes, I think it would be a good idea to implement app-boot.

PR #729, opened 2019-09-19
Changes to support embroider build for Fastboot Apps

dnalagatla:
This PR adds Support the following changes to enable embroider build:

1. Rebuild and read from index.html with updated files. This Feature will be behind the flag.
2. Added process.env.FASTBOOT_HTML_MANIFEST flag to support updating manifest from index.html.

@stefanpenner @rwjblue @ef4


lib/broccoli/fastboot-app-factory-plugin.js
```diff
@@ -0,0 +1,24 @@
...
+      annotation: 'Generate: FastBoot app-factory module',
+      persistentOutput: true
+    });
+    this.appName = options.appName;
+    this.isModuleUnification = options.isModuleUnification;
+  }
+
+  build() {
+    let outputPath = path.join(this.outputPath, 'app-factory.js');
+    let content = fastbootAppFactoryModule(this.appName, this.isModuleUnification);
```
stefanpenner: I think moduleUnification is no longer supported do we need the corresponding code?

cc @rwjblue

lib/broccoli/fastboot-app-factory-plugin.js
```diff
@@ -0,0 +1,24 @@
...
+    super([inputNode], {
+      annotation: 'Generate: FastBoot app-factory module',
+      persistentOutput: true
+    });
+    this.appName = options.appName;
+    this.isModuleUnification = options.isModuleUnification;
+  }
+
+  build() {
+    let outputPath = path.join(this.outputPath, 'app-factory.js');
```
stefanpenner: this join is not needed. fs.readFileSync(`${this.outputPath}/app-factory.js`) will suffice.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const fs = require('fs');
+const findScriptSrcs = require('find-scripts-srcs-in-document');
+const isEmbroider = require('./is-embroider');
+
+// TODO: hard coding these seem unfortunate, especially vendor-static which is non-standard
```
stefanpenner: is there an alternative, hardcoding these seems brittle/error prone.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,50 @@
...
+const FILES_TO_SKIP = /vendor|vendor-static/gi;
+
+module.exports = function updateManifestFromHtml(root) {
+  const pkg = JSON.parse(fs.readFileSync(root + '/package.json', 'UTF8'));
```
stefanpenner: should this be moved after the `isEmbroider` guard, as if that guard is truithy we will not use the `pkg`

test/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,129 @@
...
+    delete project.pkg['ember-addon'];
+    project.writeSync();
+
+    const originalJSON = project.toJSON();
+    updateManifestFromHtml(project.root + '/my-app')
+    project.readSync();
+    expect(project.toJSON()).to.eql(originalJSON);
+  });
+  // TODO: these heuristics being tested seem brittle, we should do something else.
+  it(`works correctly ass files are added/removed`,  function() {
```
stefanpenner: looks like i typo'd `ass` here, we should likely correct.

test/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,129 @@
+'use strict';
```
stefanpenner: You mentioned a scenario that https://github.com/ember-fastboot/ember-cli-fastboot/pull/727 didn't handle, but these tests appear to be the same as https://github.com/ember-fastboot/ember-cli-fastboot/pull/727 could you add the missing test, or leave a note as to the test change you added that covers that scenario?

lib/embroider/is-embroider.js
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs');
+
+module.exports = function isEmbroider(root) {
```
stefanpenner: tests please

lib/embroider/is-embroider.js
```diff
@@ -0,0 +1,10 @@
+const fs = require('fs');
```
stefanpenner: may be a missing a `'use strict';`

lib/broccoli/fastboot-app-factory-plugin.js
```diff
@@ -0,0 +1,24 @@
...
+    let content = fastbootAppFactoryModule(this.appName, this.isModuleUnification);
+    fs.writeFileSync(outputPath, content);
```
stefanpenner: should we write if the file hasn't changed?

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,50 @@
...
+
+  const indexScriptFilePaths = findScriptSrcs(indexHtml, findScriptSrcs.ignoreWithAttribute('data-embroider-ignore'))
+    // TODO: these FILES_TO_SKIP things seem over zealous
+    .filter( src => !FILES_TO_SKIP.test(src));
+
+    const appFiles = [];
+    indexScriptFilePaths.forEach(src => {
+      // Extract the file name from the src file paths and
+      // checks the file exits in the assets location on the disk.
+      const scriptFileNameRegEx = /([a-zA-Z0-9_\.\-\(\):])+(\.js)/ig;
```
stefanpenner: is the complexity in this function covered in tests?

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,50 @@
...
+      if(fileName && !appFiles.includes(assetPath) && fs.existsSync(filePath)) {
+        appFiles.push(assetPath);
+      }
+    });
+
+  // TODO: this feels super janky, we need to figure out a better approach
+  const fastbootFile = pkg.fastboot.manifest.appFiles.find(x => /-fastboot.js$/.test(x))
+
+  if (!fastbootFile) {
+    throw new Error(`ember-cli-fastbot: missing '<app-name>-fastboot.js' file`)
```
stefanpenner: `ember-cli-fastbot` -> `emer-cli-fastboot`

lib/broccoli/fastboot-app-factory-plugin.js
```diff
@@ -0,0 +1,24 @@
...
+
+const path = require('path');
+const fs = require('fs');
+const Plugin = require('broccoli-plugin');
+const fastbootAppFactoryModule = require('../utilities/fastboot-app-factory-module');
+
+module.exports = class FastBootAppFactory extends Plugin {
+  constructor(inputNode, options) {
+    super([inputNode], {
+      annotation: 'Generate: FastBoot app-factory module',
```
stefanpenner: can we update this annotation to include the exact filename?

`EmberCLIFastboot: write app-factory.js to output`

dnalagatla:
My merge with existing had some issue and this PR included some unwanted code. Cleaned up the PR

lib/embroider/update-manifest-from-html.js
dnalagatla: This Handles the following scenarios:
`%placeholder[a.js]` -> placeholder gets added to index.html as part of build process
`chunk-1234.js` - > format in which files are generated.
`some_context_path/xyx.js` ->  where application adds a context path to assets.

I have tests updated

lib/embroider/update-manifest-from-html.js
dnalagatla: I don't know if there are any alternatives to this approach other than using the attribute on script tags indicating skip for fastboot.

lib/embroider/update-manifest-from-html.js
dnalagatla: Not needed as `isEmbroider` is happening in `postBuild` hook.

dnalagatla:
> There are several todos, they should likely be resolved.
> 
> Also, there is some related conversion in another the PR that this one has replaced that needs to be handled so be sure to review comments on: see: #727 especially [#727 (comment)](https://github.com/ember-fastboot/ember-cli-fastboot/pull/727#discussion_r325960541)

I have worked on using `process.env.EMBROIDER=true` to turn on embroider build flow in fastboot. this help in keeping backward compatibility.

ef4:
I still haven't heard anyone explain why these changes aren't backward compatible. They were intended to be, and if they're not, I don't actually support merging them until we adjust to keep them backward compatible.

If people are going to insist on feature flagging this anyway, please at least pick a name other than EMBROIDER. The flag has nothing to do with embroider, it's a feature flag inside ember-cli-fastboot and it's confusing to stomp on that environment variable.

dnalagatla:
> I still haven't heard anyone explain why these changes aren't backward compatible. They were intended to be, and if they're not, I don't actually support merging them until we adjust to keep them backward compatible.
> 
> If people are going to insist on feature flagging this anyway, please at least pick a name other than EMBROIDER. The flag has nothing to do with embroider, it's a feature flag inside ember-cli-fastboot and it's confusing to stomp on that environment variable.



> I still haven't heard anyone explain why these changes aren't backward compatible. They were intended to be, and if they're not, I don't actually support merging them until we adjust to keep them backward compatible.
> 
> If people are going to insist on feature flagging this anyway, please at least pick a name other than EMBROIDER. The flag has nothing to do with embroider, it's a feature flag inside ember-cli-fastboot and it's confusing to stomp on that environment variable.

@ef4 There is a difference between the classic build and embroider build flow as a result, I was interested in having that flag: 

In Classic build flow, the following are the steps in ember-cli-fastboot build for custom app-boot -
1. Ember-cli gives a flag to turn off autoBoot by setting this property `app.options.autoRun` and is set to false by default in include hook in `ember-cli-fastboot` - https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L73
2. Then in `contentFor` hook `app-boot` type is used to define custom app-boot logic that can support Browser and Fastboot (using appFactory). - https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/index.js#L110

Embroider build doesn't support `app-boot` hook in `contentFor` to include custom app-boot logic needed for ember-cli-fastboot. - https://github.com/embroider-build/embroider/issues/183

As `app-boot` is not supported, we introduced a flag to `disable-auto-boot`, but for that flag to be effective we would have to set `app.options.autoRun` to `true`.  That change is only compatible with embroider flow but not with classic build flow of ember-cli-fastboot.

stefanpenner:
Connecting two threads together. As per https://github.com/ember-fastboot/ember-cli-fastboot/pull/727#issuecomment-534675752 it sounds like @ef4 would like you to add app-boot support. This then should keep stuff simple, and prevent hacks/future maintenance issues.

lib/embroider/update-manifest-from-html.js
stefanpenner: > other than using the attribute on script tags indicating skip for fastboot.

Yup this sounds appropriate.

dnalagatla:
> Connecting two threads together.
> 
> As per @ef4's comment [#727 (comment)](https://github.com/ember-fastboot/ember-cli-fastboot/pull/727#issuecomment-534675752) it sounds like he would like you to add app-boot support. This then should keep stuff simple, and prevent hacks/future maintenance issues.

@stefanpenner @ef4  I have created a PR to include app-boot support  - https://github.com/embroider-build/embroider/pull/335

lib/embroider/update-manifest-from-html.js
dnalagatla: Instead of hardcoding the vendor files to skip, I have added a check to verify if the file from index.html is present in `manifest.vendorFiles` and skip them. 

This will ensure we are only checking for App files and not including any vendor files.

dnalagatla:
> There are several todos, they should likely be resolved.
> 
> Also, there is some related conversion in another the PR that this one has replaced that needs to be handled so be sure to review comments on: see: #727 especially [#727 (comment)](https://github.com/ember-fastboot/ember-cli-fastboot/pull/727#discussion_r325960541)

I updated the code to address all the PRs except one related to simple-html-tokenizer, which has a path forward with a pending PR. We can fix in a separate PR.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,67 @@
...
+  // below flow as there is no package.json to update.
+  if (!fs.existsSync(root + '/package.json')) {
+    return;
+  }
+
+  const pkg = JSON.parse(fs.readFileSync(root + '/package.json', 'UTF8'));
+  const appFile = pkg.fastboot.manifest.appFiles[0];
+
+  // if app.js file exists skip reading from the index.html as
+  // it is classic build flow
```
stefanpenner: I thought the plan was to not detect classic/embroider, but to create a feature that classic and embroider could both use, put it behind a feature specific flag, to allow classic folks time to upgrade.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,67 @@
...
+  // TODO: fix simple-html-tokenizer: https://github.com/tildeio/simple-html-tokenizer/pull/71
+    .replace('<!DOCTYPE html>','');
+
+  const assetsDir = `assets/`;
+  const vendorFiles = pkg.fastboot.manifest.vendorFiles;
+
+  // Find script src which are not data-fastboot-ignore and data-embroider-ignore
+  const indexScriptFilePaths = findScriptSrcs(indexHtml, scriptToken => {
+    const { attributes } = scriptToken;
+    return attributes.find(attr => attr[0] === 'data-embroider-ignore' || 'data-fastboot-ignore');
```
stefanpenner: I believe we discussed `data-embroider-ignore` should not be part of `ember-cli-fastboot`. If a user wants to ignore both embroider and fastboot then they must include both tags.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,67 @@
...
+  }).filter( src => {
+    // skipping files if they are part of vendorFiles list
```
stefanpenner: can you briefly explain in comment, why it is safe to do so

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,67 @@
...
+    // this would cover any vendor files (vendor or vendor-static)
+    const fileName = getFileName(src);
+    return !vendorFiles.find(filePath => fileName === getFileName(filePath));
+  });
+
+  // Building appFiles list from html.
+  const appFiles = [];
+  indexScriptFilePaths.forEach(src => {
+    // Extract the file name from the src file paths and
```
stefanpenner: Instead of commenting what the code does bellow, as that could is very clear and well written, could you include the rationale and the why?

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,67 @@
...
+    const assetPath = `${assetsDir}${fileName}`;
+    const filePath = `${root}/${assetPath}`;
+    if(fileName && !appFiles.includes(assetPath) && fs.existsSync(filePath)) {
+      appFiles.push(assetPath);
+    }
+  });
+
+  const fastbootFile = pkg.fastboot.manifest.appFiles.find(x => /-fastboot.js$/.test(x))
+  if (!fastbootFile) {
+    throw new Error(`ember-cli-fastboot: missing '<app-name>-fastboot.js' file`)
```
stefanpenner: Are their common problems that can cause this? Is there something the user can do to address this issue?

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+const fs = require('fs');
+const findScriptSrcs = require('find-scripts-srcs-in-document');
+
+function getFileName(filePath) {
+  const scriptFileNameRegEx = /([a-zA-Z0-9_\.\-\(\):])+(\.js)/ig;
```
stefanpenner: This regex is complex, let's ensure it or `getFileName` are thoroughly unit tested.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,67 @@
...
+
+  // if app.js file exists skip reading from the index.html as
+  // it is classic build flow
+  if(appFile && fs.existsSync(`${root}${appFile}`)) {
+    return;
+  }
+
+  // otherwise, we must parse the index.html file to figure out what's next
+  const indexHtml = fs.readFileSync(root + '/index.html', 'UTF8')
+  // TODO: fix simple-html-tokenizer: https://github.com/tildeio/simple-html-tokenizer/pull/71
```
stefanpenner: @rwjblue how are we looking on that PR?

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,67 @@
...
+
+function getFileName(filePath) {
+  const scriptFileNameRegEx = /([a-zA-Z0-9_\.\-\(\):])+(\.js)/ig;
+  return Array.isArray(filePath.match(scriptFileNameRegEx)) ? filePath.match(scriptFileNameRegEx)[0] : '';
+}
+
+module.exports = function updateManifestFromHtml(root) {
+  // package.json is not present in root location skip
+  // below flow as there is no package.json to update.
+  if (!fs.existsSync(root + '/package.json')) {
```
stefanpenner: In what scenario, where ember-cli-fastboot is installed as an add-on, will `package.json` be missing?

lib/embroider/update-manifest-from-html.js
dnalagatla: Ah, I misunderstood ask at that time. Let me update to include a feature flag to read from index.html that supports both build flows.

lib/embroider/update-manifest-from-html.js
dnalagatla: I have covered different Filenames in the test for updating manifest from index.html. `update-manifest-from-html-test.js`

Let me see I missed anything there.

lib/embroider/update-manifest-from-html.js
dnalagatla: I will fix that.

lib/embroider/update-manifest-from-html.js
dnalagatla: The Vendor Files list is generated as part of the `ember-cli-fastboot` and during the generation of pacakge.json the list of vendor files are pulled from `this.outputPaths.vendor.js` and I think that is a proper source of truth to know all the vendor files passed to fastboot env. 

As we want to skip including vendor files in the updated list of appFile, checking src with vendorFiles made more sense to me. Please let me know if i missed anything there.

lib/embroider/update-manifest-from-html.js
dnalagatla: Actually, the only time this file will not be available if package.json generation had issues during the build. I kept this check when I pulled in your changes. 

I think we don't need that check, as the expectation after installing ember-cli-fastboot is that the proper package.json is generated during the build time. Let me remove that check.

lib/embroider/update-manifest-from-html.js
dnalagatla: THis was to cover custom app build scenarios where assets are generated in a nested directory in outputPath.

lib/embroider/update-manifest-from-html.js
dnalagatla: I fixed this to only include `data-fastboot-ignore`.

lib/embroider/update-manifest-from-html.js
dnalagatla: I removed this check, as `ember-cli-fastboot` will always generate the package.json with `<appname>-fastboot.js`.

lib/embroider/update-manifest-from-html.js
dnalagatla: I covered following FileName scenarios in my testing
```
 <script src="%PLACEHOLDER[c.js]%"></script>
  <script src="/some_context_path/e.js"></script>
  <script src="chunk-12.js"></script>
  <script src="chunk-12.test.js"></script>
```

lib/embroider/update-manifest-from-html.js
dnalagatla: I Fixed this. Added a feature (non-embroider) specific to make it work for any build flow.

Thank you.

stefanpenner:
thoughts on renaming UPDATE_MANIFEST_FROM_HTML  to FASTBOOT_UPDATE_MANIFEST_FROM_HTML

lib/embroider/update-manifest-from-html.js
stefanpenner: Also, could you briefly explain what the purpose of the regexp as well?

lib/embroider/update-manifest-from-html.js
stefanpenner: I believe this then should cause an error, rather than silently failing for such apps.

dnalagatla:
> FASTBOOT_UPDATE_MANIFEST_FROM_HTML

that makes sense. I will update that.

lib/embroider/update-manifest-from-html.js
dnalagatla: I added the rationale behind this parsing.

lib/embroider/update-manifest-from-html.js
dnalagatla: I removed this check for now. As we discussed, for multiple apps usecase, we can provide a broccoli plugin for updating index.html. I would like to cover that as separate task.

index.js
```diff
@@ -346,7 +347,13 @@ module.exports = {
   },
 
   postBuild(result) {
-    this._updateAppFilesInManifest(result.directory);
+    // Need to update manifest from index.html.
+    // Set environment variable `UPDATE_MANIFEST_FROM_HTML` to `true`
+    // Usage `UPDATE_MANIFEST_FROM_HTML=true ember s`
+    if(process.env.FASTBOOT_UPDATE_MANIFEST_FROM_HTML === 'true') {
```
kratiahuja: When will this be set to true? Is it only for embroider builds that this flag is true?

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,53 @@
...
+
+const fs = require('fs');
+const findScriptSrcs = require('find-scripts-srcs-in-document');
+const getScriptFileName = require('../utilities/get-script-file-name');
+
+module.exports = function updateManifestFromHtml(root) {
+  const pkg = JSON.parse(fs.readFileSync(root + '/package.json', 'UTF8'));
+
+  // otherwise, we must parse the index.html file to figure out what's next
+  const indexHtml = fs.readFileSync(root + '/index.html', 'UTF8')
```
kratiahuja: IIRC, doesn't ember-cli provide an API the default index.html name and have the app call it something else? If that is the case, this will throw an error.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+const fs = require('fs');
+const findScriptSrcs = require('find-scripts-srcs-in-document');
```
kratiahuja: Did we consider leveraging `jsdom` for this?

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,53 @@
...
+  const appFiles = [];
+  indexScriptFilePaths.forEach(src => {
+    // Extract the file name from the src file paths and
+    // checks the file exits in the assets location on the disk.
+    const fileName = getScriptFileName(src);
+    const assetPath = `${assetsDir}${fileName}`;
+    const filePath = `${root}/${assetPath}`;
+    if(fileName && !appFiles.includes(assetPath) && fs.existsSync(filePath)) {
+      appFiles.push(assetPath);
+    }
```
kratiahuja: Wouldn't this include other script tags that may not be app files? Example I may add a google analytics script tag in my index.html. Seems like this might also be included when it shouldn't because we aren't checking for `data-fastboot-ignore` here. 

Also how would this work for LI apps that will have filename wrapped in spark placeholders?

lib/utilities/get-script-file-name.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+module.exports = function getScriptFileName(filePath) {
+  const scriptFileNameRegEx = /([a-zA-Z0-9_\.\-\(\):])+(\.js)/ig;
+  return Array.isArray(filePath.match(scriptFileNameRegEx)) ? filePath.match(scriptFileNameRegEx)[0] : '';
+}
```
kratiahuja: I think if we use `jsdom` we can directly use APIs like `getAttribute('src')` which might eliminate the need for the regex.

lib/utilities/get-script-file-name.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
+
+module.exports = function getScriptFileName(filePath) {
+  const scriptFileNameRegEx = /([a-zA-Z0-9_\.\-\(\):])+(\.js)/ig;
```
kratiahuja: Shouldn't this be a constant outside the function so it is only compiled once?

lib/embroider/update-manifest-from-html.js
kratiahuja: For the second question, I see you have added tests already. So I think that answers my second question.

index.js
stefanpenner: @kratiahuja this will also be true for non embroider builds using this new "treat index.html" as the entry point feature.  If this feature turns out to work well, we would like to explore it being the default.

lib/embroider/update-manifest-from-html.js
stefanpenner: @kratiahuja in discussions with @rwjblue we decided not to add another parser, rather use the same parser embers template system already uses: https://github.com/tildeio/simple-html-tokenizer. Future work, may want to explore using simple-html-tokenizer more often, that way we avoid having many different DOM parsing implementations around. Fully dropping JSDom would require more effort, so we should consider it case/case.

lib/utilities/get-script-file-name.js
stefanpenner: If memory serves regexe's with `/g` have state

lib/utilities/get-script-file-name.js
stefanpenner: see above comment

stefanpenner:
questions for @rwjblue  / @ef4 / @dnalagatla since this is an experimental feature, should we release this behind a flag, or should we release fork of this project?

ef4:
I think you should use a fork for the present.

I only suggested making the change here because there's an opportunity to simplify the architecture. But if there isn't interest / energy / consensus to change the architecture of ember-cli-fastboot (to no longer need a manifest), then having this special mode is *adding* complexity, and at that point I'd rather just paper over the problem from within embroider itself.

lib/utilities/get-script-file-name.js
```diff
@@ -0,0 +1,6 @@
...
+  return Array.isArray(filePath.match(scriptFileNameRegEx)) ? filePath.match(scriptFileNameRegEx)[0] : '';
```
rwjblue: Can you explain why we need this regexp matching?

Also, we need to simplify this:

```suggestion
  let match = filePath.match(scriptFileNameRegEx);
  return match ? match[0] : '';
```

index.js
kratiahuja: YEah I think that would definitely help people understand the changes. I think we probably also need to update the readme for the same so folks can opt-in.

lib/embroider/update-manifest-from-html.js
kratiahuja: Aah ok, thanks for the context.

dnalagatla:
> Can you update the title and description to be a current summary of the PR?

Hi @rwjblue, I made the necessary changes to the Title and description.

lib/utilities/get-script-file-name.js
dnalagatla: That Makes sense. I updated the code.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,53 @@
...
+module.exports = function updateManifestFromHtml(root) {
+  const pkg = JSON.parse(fs.readFileSync(root + '/package.json', 'UTF8'));
+
+  // otherwise, we must parse the index.html file to figure out what's next
+  const indexHtml = fs.readFileSync(root + '/index.html', 'UTF8')
+  // TODO: fix simple-html-tokenizer: https://github.com/tildeio/simple-html-tokenizer/pull/71
+    .replace('<!DOCTYPE html>','');
+
+  const assetsDir = `assets/`;
+  const vendorFiles = pkg.fastboot.manifest.vendorFiles;
```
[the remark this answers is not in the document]
dnalagatla: I made changes to be defensive by falling back to an empty array if that property is not defined. In build flow of `ember-cli-fastboot-addon` though the config generated will always include the vendor files list.

SparshithNR:
@dnalagatla can we merge this PR after resolving the conflict?
@stefanpenner @rwjblue @kratiahuja @ef4  Looks like @dnalagatla addressed all the comments in the PR.

dnalagatla:
> @dnalagatla can we merge this PR after resolving the conflict?
> @stefanpenner @rwjblue @kratiahuja @ef4 Looks like @dnalagatla addressed all the comments in the PR.

@SparshithNR  Thank you, will take a look into resolving the conflicts.

index.js
dnalagatla: Added a reed me section for this flag.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,59 @@
...
+  // otherwise, we must parse the index.html file to figure out what's next
+  const indexHtml = fs.readFileSync(root + '/' + htmlFileName, 'UTF8')
+  // TODO: fix simple-html-tokenizer: https://github.com/tildeio/simple-html-tokenizer/pull/71
+    .replace('<!DOCTYPE html>','');
+
+  const assetsDir = `assets/`;
+  const vendorFiles = pkg.fastboot.manifest.vendorFiles || [];
+
+  // Find script src which are not data-fastboot-ignore
+  const indexScriptFilePaths = findScriptSrcs(indexHtml, findScriptSrcs.ignoreWithAttribute('data-fastboot-ignore'))
```
kratiahuja: Who will be adding `data-fastboot-ignore` attribute?

test/update-manifest-from-html-test.js
```diff
@@ -0,0 +1,128 @@
...
+    files['a.js'] = '';
+    files['b.js'] = '';
+    files['c.js'] = '';
+    files['d.js'] = '';
+    files['e.js'] = '';
+    files['chunk-12.js'] = '';
+    files['chunk-12.test.js'] = '';
+
+    fixturify.writeSync(project.root + '/my-app/assets', files);
+    updateManifestFromHtml(project.root + '/my-app')
```
kratiahuja: Missing semi-colon

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,59 @@
...
+    // vendor files as they are already included in manifest.
+    // Checking if the script file from index.html
+    // is in the the list of vendor files in package.json and skip them.
+    const fileName = getScriptFileName(src);
+    return !vendorFiles.find(filePath => fileName === getScriptFileName(filePath));
+  });
+
+  /**
+   * Reading app files by parsing index.html so that FastBoot can load assets that
+   * are same as what is generategd as part of the build flow and updated in
```
kratiahuja: generated

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,59 @@
...
+   *
+   * Example: Embroider build flow upon WebPack as stage 3 processing in the build
+   * updates app.js to multiple chunks and updates index.html. Parsing from index.html
+   * allows loading same set of app files in FastBoot sandbox.
+   *
+   */
+  const appFiles = new Set();
+  indexScriptFilePaths.forEach(src => {
+    // Extract the file name from the src file paths and
+    // checks the file exits in the assets location on the disk.
```
kratiahuja: exits -> exists

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,59 @@
...
+    const filePath = `${root}/${assetPath}`;
+    if(fileName && fs.existsSync(filePath)) {
+      appFiles.add(assetPath);
+    }
+  });
+
+  const fastbootFile = pkg.fastboot.manifest.appFiles.find(x => /-fastboot.js$/.test(x));
+  appFiles.add(fastbootFile);
+  pkg.fastboot.manifest.appFiles = Array.from(appFiles);
+  fs.writeFileSync(`${root}/package.json`, JSON.stringify(pkg, null, 2));
```
kratiahuja: We should add a comment why are we doing this and how is it different than `updateFastBootManifest` hook.

index.js
```diff
@@ -346,6 +347,13 @@ module.exports = {
   },
 
   postBuild(result) {
+    // Need to update manifest from html file.
+    // Set environment variable `FASTBOOT_HTML_MANIFEST` to `true`
+    // Usage `FASTBOOT_HTML_MANIFEST=true ember s`
+    if(process.env.FASTBOOT_HTML_MANIFEST === 'true') {
+      updateManifestFromHtml(result.directory);
+    }
```
kratiahuja: In order for this to work correctly internally, we need to make sure our placeholder addon runs before fastboot.

test/update-manifest-from-html-test.js
```diff
@@ -0,0 +1,128 @@
...
+
+  PROJECT.pkg.fastboot = {
+    manifest: {
+      htmlFile: 'index.html',
+      appFiles: ['assets/app.js', 'assets/my-app-fastboot.js'],
+      vendorFiles:['assets/vendor.js', 'assets/vendor-static.js']
+    }
+  };
+
+  it(`works correctly as files are added/removed`,  function() {
```
kratiahuja: This file is testing lot of testcases into one. My suggestion is we should split this test.

lib/utilities/fastboot-app-boot.js
```diff
@@ -7,7 +7,7 @@ module.exports = function fastbootAppBoot(prefix, configAppAsString, isModuleUni
   return [
     "",
     "if (typeof FastBoot === 'undefined') {",
-    "  if (!runningTests) {",
+    "  if (typeof runningTests === 'undefined' || !runningTests) {",
```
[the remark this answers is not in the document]
kratiahuja: Can we add a comment for the same?

lib/embroider/update-manifest-from-html.js
SparshithNR: https://github.com/ember-fastboot/ember-cli-fastboot/pull/729/files#diff-b3e864678d174a0e2890fc5bbe5939bdR33-R42 
Explains why we need to do this.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,53 @@
...
+    // checks the file exits in the assets location on the disk.
+    const fileName = getScriptFileName(src);
+    const assetPath = `${assetsDir}${fileName}`;
+    const filePath = `${root}/${assetPath}`;
+    if(fileName && !appFiles.includes(assetPath) && fs.existsSync(filePath)) {
+      appFiles.push(assetPath);
+    }
+  });
+
+  const fastbootFile = pkg.fastboot.manifest.appFiles.find(x => /-fastboot.js$/.test(x));
```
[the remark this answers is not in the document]
SparshithNR: https://github.com/ember-fastboot/ember-cli-fastboot/blob/9ace0cd548a155d2ac92616ca0eb9acb97088423/lib/embroider/update-manifest-from-html.js#L57
Made this change separate in #753  as carrying changes from this.

lib/utilities/fastboot-app-boot.js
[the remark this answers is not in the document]
SparshithNR: https://github.com/ember-fastboot/ember-cli-fastboot/blob/9ace0cd548a155d2ac92616ca0eb9acb97088423/lib/utilities/fastboot-app-boot.js#L10

lib/embroider/update-manifest-from-html.js
SparshithNR: Added this info to README and in code as well 
https://github.com/ember-fastboot/ember-cli-fastboot/blob/9ace0cd548a155d2ac92616ca0eb9acb97088423/README.md#update-manifest-from-html
https://github.com/ember-fastboot/ember-cli-fastboot/blob/9ace0cd548a155d2ac92616ca0eb9acb97088423/lib/embroider/update-manifest-from-html.js#L20

Issue #730, opened 2019-09-20
Build breaks for a custom app on 2.2.1


sangm:
Trying to build our app on `2.2.1`

We are seeing this exception:

```
Restarting BPR processes (pkill -9 -f "renderer.js")
(node:86010) UnhandledPromiseRejectionWarning: Error: Cannot find module '/var/folders/h2/c4_3_zwd0v3298rc7jyl_ljh000gv4/T/broccoli-86010PiEIdOkWFbBv/out-13001-broccoli_merge_trees_merge_build_trees/package.json'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:581:15)
    at Function.Module._load (internal/modules/cjs/loader.js:507:25)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Class._updateAppFilesInManifest (foo-project/node_modules/ember-cli-fastboot/index.js:384:17)
    at Class.postBuild (foo-project/node_modules/ember-cli-fastboot/index.js:349:10)
    at addonPromises.project.addons.reduce (foo-project/node_modules/ember-cli/lib/models/builder.js:182:28)
    at Array.reduce (<anonymous>)
    at Builder.processAddonBuildSteps (foo-project/node_modules/ember-cli/lib/models/builder.js:179:43)
    at tryCatcher (foo-project/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:323:19)
    at invokeCallback (foo-project/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:495:31)
    at publish (foo-project/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:481:7)
    at flush (foo-project/node_modules/ember-cli/node_modules/rsvp/dist/rsvp.js:2402:5)
    at process._tickCallback (internal/process/next_tick.js:61:11)
(node:86010) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:86010) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
[...]

PR #737, opened 2019-10-29
Fix custom app build issue #730

dnalagatla:
Fixed issue #730 
Revert "cheerio is used outside of tests so must be a dependency not a"

This reverts commit 097ad839c6f494988b5ef18e7dddaf53db313470.

Revert "Added support to update manifest app files from index.html"

This reverts commit 0a12a6f226f037e83ecb256c18519b47269e9dc0.

@rwjblue @kratiahuja

### DIFF
--- a/index.js
+++ b/index.js
@@ -346,7 +346,6 @@ module.exports = {
   },
 
   postBuild(result) {
-    this._updateAppFilesInManifest(result.directory);
     if (this.fastboot) {
       // should we reload fastboot if there are only css changes? Seems it maynot be needed.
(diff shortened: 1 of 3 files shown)


ef4:
This is still unreleased, so the broken feature is floating around breaking things.

kratiahuja:
@ef4 do you want a release? I can release a new version.

ef4:
Yes, thanks that would be good.

kratiahuja:
@ef4  sorry for the delay. `ember-cli-fastboot@2.2.2` is released

PR #749, opened 2020-02-03
Updated fastboot service to not throw error when request object is not passed as visit options

dnalagatla:
This change handles the scenario when the fastboot library's visit options don't include request property. This happens when fastboot library is used without fastboot-middleware and metadata is used for handling request context.  

* Added unit tests for above scenario.


dnalagatla:
cc: @kratiahuja  @rwjblue

addon/services/fastboot.js
```diff
@@ -9,6 +9,10 @@ import Service from '@ember/service';
 const RequestObject = EObject.extend({
   init() {
     this._super(...arguments);
+    // Adding a defensive check to support scenario when visit options doesn't include request object
+    if(!this.request) {
+      return;
+    }
```
kratiahuja: Instead of checkinf for this in the object, could we abstract these objects into utils and then simply override it to provide a default empty object.

dnalagatla: Hi Krati,

Based on your suggestions I abstracted request-object and shoebox to separate utils module. But to support overriding of those modules I had to move/create the following files in app folder instead of having them in addon location.

services/fastboot.js
utils/request-object.js
utils/shoebox.js

This allows overriding of the above objects with custom objects created by the application. In general I don't see any issues with this change, but I am not sure if this a general pattern followed during addon development.

dnalagatla: Discussed offline with @kratiahuja. Discarding this PR.

dnalagatla:
Discussed offline with @kratiahuja. Discarding this PR.

PR #751, opened 2020-02-15
Add option for custom FastBoot implementation

Ravenstine:
I think it's good that ember-cli-fastboot comes with the FastBoot library out of the box, but it would be nice to override that and use your own implementation of FastBoot.

The case where this would be most useful is when someone would want to use a fork of FastBoot, whether it's their own or someone else's.  My motivation for adding this option to ember-cli-fastboot so that I can use my [FastBoot fork](https://github.com/ravenstine/powerboot) without having to use a hackey postinstall script to make it happen with a symbolic link.

In short, this allows specifying a custom FastBoot class in your `ember-cli-build.js` file:

```js
const FastBoot = require('my-fastboot-fork');
const app = new EmberApp({
  fastboot: {
    implementation: FastBoot
  }
});
```

I'm sure most people may not need this, but it's something I would like and I figured I'd put it out there in case anyone thinks it makes sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -551,6 +551,20 @@ fastbootConfigTree() {
 
 The above configuration will be available in Node via the `FastBoot.config()` function. Therefore, in order to get the above config, the addon/app can call `FastBoot.config('<engine-name>')`.
 
+## Custom FastBoot Implementation
+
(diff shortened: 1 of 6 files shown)


kratiahuja:
Could you please define what your usecase is to fork `fastboot`? Is there something missing that is better built within `fastboot`?

PR #752, opened 2020-02-18
Add support for custom configuration

bobisjan:
This adds support for reading FastBoot configuration from `config/fastboot.js` (similar to how other add-ons like `ember-intl` works).

It should be possible to share this file with production server `fastboot-app-server`.

**TODOs**

- [x] Update `README.md`
- [ ] Add blueprint


index.js
```diff
@@ -79,6 +79,8 @@ module.exports = {
     this._appRegistry = app.registry;
     this._name = app.name;
 
+    this.options = this._optionsFor(app.env, app.project);
```
kratiahuja: Nitpick: Maybe better to name this `this.fastbootOptions` for clarity.

test/fastboot-config-test.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+const expect = require('chai').use(require('chai-string')).expect;
+const RSVP = require('rsvp');
+const request = RSVP.denodeify(require('request'));
+
+const AddonTestApp = require('ember-cli-addon-tests').AddonTestApp;
+
+describe('FastBoot config', function() {
+  this.timeout(400000);
```
kratiahuja: Any reason for such a high timeout?

index.js
```diff
@@ -373,8 +376,17 @@ module.exports = {
...
     return (typeof this.project.isModuleUnification === 'function') && this.project.isModuleUnification();
+  },
+
+  _optionsFor(environment, project) {
+    const configPath = path.join(path.dirname(project.configPath()), 'fastboot.js');
+
+    if (fs.existsSync(configPath)) {
+      return require(configPath)(environment);
+    }
+    return {};
```
kratiahuja: Can you add comments for this function?

Also can an addon provide additional options that can be merged with what the host app provides? Should we let addons also provide a way to put configs?

test/fastboot-config-test.js
bobisjan: Just copy&paste from another test, I'll remove.

index.js
bobisjan: > Also can an addon provide additional options that can be merged with what the host app provides? Should we let addons also provide a way to put configs?

Good questions, and TBH I don't know 🙈, my primary motivation for this is to avoid fork of `ember-cli-fastboot` due to custom globals requirements https://github.com/zonkyio/ember-cli-fastboot/blob/for-zonky-3-0/index.js#L323. These globals are keys (private, which should not leak to browser) for application, not sure if this add-on can provide on it's own.

test/fastboot-config-test.js
bobisjan: I had to put back the timeout due to failing tests, not sure if the number is too high.

index.js
kratiahuja: Can you add a TODO for now so we know this is a limitation and when there is a usecase we can add it.

bobisjan: Done ✅.

bobisjan:
> Please update the README

Done ✅, the section uses option and link for `fastboot#2.x`, should be updated when upgrading to `fastboot#3.x` in the future.

README.md
```diff
@@ -60,6 +60,24 @@ visit your app at `http://localhost:4200/?fastboot=false`. If you want to turn o
 
 You can even disable serving fastboot with `ember serve` using an environment flag: `FASTBOOT_DISABLED=true ember serve`. If you have disabled building fastboot assets using the same flag as described [here](https://github.com/ember-fastboot/ember-cli-fastboot#double-build-times-and-no-incremental-builds), remember to also disable serving fastboot assets when using `ember serve`.
 
+### FastBoot Configuration
+
+When running locally using `ember serve` you can pass options into FastBoot instance via `config/fastboot.js` file.
```
kratiahuja: Could you please update this to mention this file should only be define in the application? Want to make sure this isn't confusing for addon authors.

README.md
```diff
@@ -60,6 +60,24 @@ visit your app at `http://localhost:4200/?fastboot=false`. If you want to turn o
...
+  let myGlobal = environment === 'production' ? process.env.MY_GLOBAL : 'testing';
+
+  return {
+    sandboxGlobals: {
+      myGlobal;
+    }
+  };
+}
+```
+
```
kratiahuja: The current code path will not merge the `myGlobal` in production I believe. So do you need to define production? It is primarily for `ember s` only.

bobisjan: Yeah, this might be little bit confusing - the environment is provided by `ember-cli`, hence the `ember s -prod` will also set environment to `production`.

Where this file will not work is when used with `fastboot-app-server`, but this is different setup - should be the README updated with mention about `fastboot-app-server` (the example usage can also be provided)?

kratiahuja: Yes please it may be good to specific it as it will help consumers not assume this will work when using `fastboot-app-server`.

kratiahuja:
This is good to go IMO after the minor update to readme about `fastboot-app-server`.

README.md
bobisjan: 👍, I've added section about FastBoot App Server below, you can take a look.

kratiahuja:
Thanks for addressing the comments. Merging this.

PR #753, opened 2020-02-20
Changes required for Embroider build to support Fastboot apps

SparshithNR:
This PR address the review comments from #729


dnalagatla:
Thank you @SparshithNR for working on fixing those new review comments. The changes look good to me.

README.md
```diff
@@ -436,6 +436,20 @@ use the rehydration builder and your application will be using rehydration.
...
+
+```
+FASTBOOT_HTML_MANIFEST=true
+```
+
+If we decide certain scirpt tags to be ignored in fastboot and shouldn't be loaded into fastboot sandbox, we need to use `data-fastboot-ignore` in script tag.
+Example:
+```
+<script src="fastboot-ignore.js" data-fastboot-ignore></script>
+```
```
kratiahuja: Please provide a concrete example of when this usecase is needed? Otherwise why are we providing this API?

test/update-manifest-from-html-test.js
```diff
@@ -0,0 +1,134 @@
...
+  const files = {};
+  files['a.js'] = '';
+  files['b.js'] = '';
+  files['c.js'] = '';
+  files['d.js'] = '';
+  files['e.js'] = '';
+  files['chunk-12.js'] = '';
+  files['chunk-12.test.js'] = '';
+
+  it(`works correctly as files are added/removed`,  function() {
```
kratiahuja: I didn't see a test case for `data-fastboot-ignore` being tested..

SparshithNR: https://github.com/ember-fastboot/ember-cli-fastboot/blob/9ace0cd548a155d2ac92616ca0eb9acb97088423/test/update-manifest-from-html-test.js#L18

README.md
dnalagatla: I think the goal here was to ignore Js files such are google-analytics as they are only needed for Browser and shouldn't be included in Fastboot Manifest.

dnalagatla: We should provide that as an example.

lib/embroider/update-manifest-from-html.js
```diff
@@ -0,0 +1,61 @@
...
+  // otherwise, we must parse the index.html file to figure out what's next
+  const indexHtml = fs.readFileSync(root + '/' + htmlFileName, 'UTF8')
+  // TODO: fix simple-html-tokenizer: https://github.com/tildeio/simple-html-tokenizer/pull/71
+    .replace('<!DOCTYPE html>','');
+
+  const assetsDir = `assets/`;
+  const vendorFiles = pkg.fastboot.manifest.vendorFiles || [];
+
+  // Find script src which are not data-fastboot-ignore
+  // data-fastboot-ignore in the script tag is added by the developer
```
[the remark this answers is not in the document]
SparshithNR: I have added an example in README. 
https://github.com/ember-fastboot/ember-cli-fastboot/pull/753/commits/29aefe0b8dfa8af62cb22fea671386a3f085e592#diff-04c6e90faac2675aa89e2176d2eec7d8R450

test/update-manifest-from-html-test.js
SparshithNR: https://github.com/ember-fastboot/ember-cli-fastboot/blob/29aefe0b8dfa8af62cb22fea671386a3f085e592/test/update-manifest-from-html-test.js#L63-L69

ef4:
This is still an important issue, but after much more end-to-end testing of full apps with fastboot & embroider I don't actually think this is the right solution.

Instead of mangling the build output in `postBuild` -- a step I can't really guarantee will always even exists -- the core code in this PR belongs in https://github.com/ember-fastboot/fastboot instead, where it can discover the right assets when the Fastboot instance is reading from `distPath`.

Also, the earlier reverted version of this feature is still present in the latest release because no release has been cut since it was reverted.

kiwiupover:
@dnalagatla @ef4 and @rwjblue what needs to happen to this PR before we can cut an embroider version of `ember-cli-fastboot`

ef4:
This PR is superseded, the equivalent changes were already made in fastboot itself.

We probably do need a release of ember-cli-fastboot that upgrades to that newer fastboot.

There is also work to do here to make ember-cli-fastboot natively emit the new format accepted by fastboot, although that is not a blocker because fastboot accepts all older formats too.

PR #767, opened 2020-07-08
Monorepo

kiwiupover:
Move ember-cli-fastboot addon to packages/ember-cli-fastboot

This PR is the base for moving tests to a monorepo structure following the discussion here. 

https://github.com/ember-fastboot/ember-cli-fastboot/issues/765

PR #768, opened 2020-07-10
Adding monorepo tests apps

kiwiupover:
### The problems

- The current test setup needs to run against npm 4 because we are swapping node modules at build time. Both `npm` 5+ and `yarn` expect node modules to be referenced from a repository or from disk somewhere. Pre `npm` 5+ node modules in present in the `node_modules` would be included when `npm install` is run.  

- [ember-cli-addon-tests](https://github.com/tomdale/ember-cli-addon-tests) builds uses ember-data@3.8.0 ember-cli-fastboot needs to tests against a wide rage of dependancies

- We need to add tests for an embroider application. 

### The case for a monorepo
We add `test-package` applications so we can tests again each of the scenarios `ember-cli-fastboot` supports. We would add an `embroider` application and run tests against that app.  A number of addons have has success with this approach `ember-auto-import` being one. 

#### Proposed test structure
I propose that we have singular focused tests and a `test-package` each of those focuses. 
- A classic app with tests for routing and the build tests
- An embroider app with routing and build tests
- Custom apps for a customization we support. Finger printing, output files names, custom html

An alternative could be we have a single custom app with all the customization tested in that application. I don't like this idea because it will be hard to fellow contributors to follow the customizations . 

### Use github actions
This is a no brainer. Test are much faster. 

Here is the same PR on my repo running github actions. https://github.com/kiwiupover/ember-cli-fastboot/pull/8 
I don't know why github actions don't run for a PR from a forked repo. 

This PR branches from the monoropo PR #767 and should be merged following the merge of that PR. 

I broke down [package-json-test.js](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/test/package-json-test.js) into separate tests with corresponding test-package applications 

Adding test-packages for  
- basic-app
- fingerprinted app
- custom html app
- custom output app

@rwjblue @kratiahuja, @SparshithNR and @mansona I would love to hear your thoughts about this approach.


kiwiupover:
Closing this PR in favor of smaller ones.

PR #771, opened 2020-07-23
Update README.md

jad359:
Removed "Introduction to Ember FastBoot" video that is no longer available on Vimeo.
Updated reference of "Ember community slack" to "Ember Community Discord".


mansona:
Any reason why a change to the readme would break the build? 🙈  cc @rwjblue @kiwiupover ?

rwjblue:
@mansona - No, I don't think it should have but its also fine to land so....

PR #776, opened 2020-07-31
Move asset rewriting test to test-packages/basic-app

kiwiupover:
This PR also replaces ember-cli-fastboot/test/asset-rewriting-test.js to basic app

@rwjblue


test-packages/basic-app/test/asset-rewriting-test.js
```diff
@@ -0,0 +1,21 @@
+const chai = require('chai');
+const expect = chai.expect;
+const glob = require("glob");
+const execa = require("execa");
+
+chai.use(require('chai-fs'));
+
+describe.only('rewriting HTML', function () {
```
mansona, line 8:
```suggestion
describe('rewriting HTML', function () {
```
😉 

I wonder if we can enable a lint rule to catch these?

kiwiupover: 😞  thanks heaps @mansona

PR #803, opened 2020-11-13
Dave/custom fastboot app

kiwiupover:
Creating the second test ember application. This app will be for customizations.


packages/ember-cli-fastboot/.eslintrc.js
```diff
@@ -15,6 +15,8 @@ module.exports = {
     browser: true
   },
   rules: {
+    'ember/no-get': 'warn',
```
kiwiupover, line 18: Adding a new ember app to the test-packages has updated `ember-cli-eslint` which has added new linting rules.

kiwiupover (commented):
@rwjblue this is ready for review.

PR #805, opened 2020-11-28
Setup Monorepo

xg-wang:
## What it does

This PR sets up the monorepo for several repositories under `ember-fastboot` org, and internal test packages for easier testing across several repos:

- `packages/fastboot`: a low-level tool for anything relying on rendering Ember app in node environment (e.g., custom fastboot-server, prember)
- `packages/ember-cli-fastboot`: required for all SSR ember app
- `packages/fastboot-express-middleware`: low-level tool for building custom fastboot-server
- `packages/fastboot-app-server`: sane default for serving production app
- `test-packages/*`: internal ember apps for acceptance & e2e testing, not published

All packages to be published are bumped to `3.1.2`, which is the largest among the packages' versions.

Also converting several `fastboot` repo tests that rely on `ember-cli-fastboot` for building ember app into `test-packages/integration-tests/test/basic-test.js`.

`ember-cli-fastboot` is already using `release-it-yarn-workspace` to release `ember-cli-fastboot`; we can continue to use the release flow to release others.

## Context

[`ember-fastboot` org](https://github.com/ember-fastboot) has multiple packages useful for server-side rendered ember apps.
Although these packages can be used together with different versions combination (e.g., https://github.com/ember-fastboot/fastboot/blob/v3.1.2/src/fastboot-schema.js is explicitly designed for compatibility), it is usually desired to keep them in sync from maintainers' perspective and use a combination of versions known good from users'.

Currently, when bumping one package, several PRs need to cascade for other packages; it becomes even trickier when a fastboot contributor needs to make a change or write a test that interacts with different packages.

To merge these related packages into a monorepo, we can make maintaining and contributing to fastboot much easier.

The cons for doing monorepo is losing the flexibility to release different versions for each package and increasing the coupling, but they are not necessarily bad IMO.

## Checklist

The monorepo setup needs to ensure a couple of things:
- [x] link packages in dev/test
- [x] ci pass
- [x] move some fastboot tests to integration-tests for demonstration ([basic-test.js](https://github.com/xg-wang/ember-cli-fastboot/blob/monorepo/test-packages/integration-tests/test/basic-test.js))
- [x] update packages/*/readme.md
- [x] release setup
- [x] keep all the git history of the different packages
- (need to verify after merge) make sure the list of contributors on GitHub is a superset

## Next steps

- [ ] After this PR is merged, move issue/pr from other repos then archive (We can only move manually/individually now 😥)
- [ ] Identify and fix any incompatibilities
[...]

PR #806, opened 2021-01-06
replace term whitelist with allowlist

mcfiredrill:
I followed the RFC from awhile back that has since been merged, so
hopefully this can be changed here in ember-cli-fastboot as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,13 +92,13 @@ server.start();
 
 ## Using Node/npm Dependencies
 
-### Whitelisting Packages
+### Allowlisting Packages
 
 When your app is running in FastBoot, it may need to use Node packages
 to replace features that are available only in the browser.
 
 For security reasons, your Ember app running in FastBoot can only access
-packages that you have explicitly whitelisted.
+packages that you have explicitly listed as allowed.
 
 To allow your app to require a package, add it to the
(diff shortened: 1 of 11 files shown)


kiwiupover:
@mcfiredrill thank you so much for your PR. I really want to make sure that we land it soon. 

To fix the test failures we there needs to be an update to the underlying `fastboot` module. You can see that here: https://github.com/ember-fastboot/fastboot/search?q=whitelist

We are in the process of pulling all the fastboot packages together into [one repo](https://github.com/ember-fastboot/ember-cli-fastboot/pull/805) which will make this PR much easier to land. Once #805 is merge then this PR will be very easy to make the change across all the fastboot packages.

mcfiredrill:
Great thank you!  Sorry I didn't have time to look into the test failures yet. 

> To fix the test failures we there needs to be an update to the underlying fastboot module. You can see that here: https://github.com/ember-fastboot/fastboot/search?q=whitelist

Makes sense.
Is it worth making a PR to ember-fastboot to fix , or maybe it's better to wait until the monorepo is setup?

kiwiupover:
@mcfiredrill Hello we just merged the monorepo branch please feel free to update your PR. 

We would like to use the name `allowedHosts` going forward. Does that make sense? 

Thanks for your help here.

PR #808, opened 2021-01-22
Fix treeForFastBoot when project is an addon

### LINKED ISSUE #807: treeForFastBoot is not handled correctly when project is an addon
When build the dummy app of an addon, the `treeForFastBoot` hook is handled differently then when the addon is included in a real app. 

First the `treeForFastBoot` hook is called as expected for the addon [here](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/packages/ember-cli-fastboot/index.js#L184). Let's asume the hook return a modified version of the addons `/fastboot` folder through some broccoli transforms.But then on the [following lines](https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/packages/ember-cli-fastboot/index.js#L185-L190), the original folder (`this.project.root + '/fastboot'`) is merged on top of that, basically overriding all the changes that the `treeForFastBoot` hook provided.

I believe the latter part is correct for an *app*, but when the project is an *addon*  this should not happen, as the addon's fastboot tree is already handled before.

simonihmig:
Fixes #807

Tested locally in an addon.

### DIFF
--- a/packages/ember-cli-fastboot/index.js
+++ b/packages/ember-cli-fastboot/index.js
@@ -184,7 +184,8 @@ module.exports = {
     this._processAddons(this.project.addons, fastbootTrees);
     // check the parent containing the fastboot directory
     const projectFastbootPath = path.join(this.project.root, 'fastboot');
-    if (this.existsSync(projectFastbootPath)) {
+    // ignore the project's fastboot folder if we are an addon, as that is already handled above
+    if (!this.project.isEmberCLIAddon() && this.existsSync(projectFastbootPath)) {
       let fastbootTree = this.treeGenerator(projectFastbootPath);
       fastbootTrees.push(fastbootTree);
     }


packages/ember-cli-fastboot/index.js
```diff
@@ -184,7 +184,8 @@ module.exports = {
     this._processAddons(this.project.addons, fastbootTrees);
     // check the parent containing the fastboot directory
     const projectFastbootPath = path.join(this.project.root, 'fastboot');
-    if (this.existsSync(projectFastbootPath)) {
+    // ignore the project's fastboot folder if we are an addon, as that is already handled above
+    if (!this.project.isEmberCLIAddon() && this.existsSync(projectFastbootPath)) {
```
kiwiupover, line 188: @simonihmig it would be great to add a test for this. Seems like we will need to add a `test-package` app that consumes an addon that uses `ember-cli-fastboot` 

#805 is going to change this repos structure over the next week. We may need to rebase this PR then we can get it merged.

kiwiupover (approved):

kiwiupover:
@simonihmig thanks for you PR

rwjblue (approved):
Chatted with @xg-wang and @kiwiupover in todays FastBoot meeting, @kiwiupover is going to work on adding a full dummy app test project so that this code is properly undertest. But we are going to merge this in while he works on that, since the code seems correct.

PR #814, opened 2021-03-16
Throw a helpful error when people use `isFastboot` instead of `isFastBoot`

bertdeblock:
https://discord.com/channels/480462759797063690/485861149821239298/821177803721801778

### DIFF
--- a/packages/ember-cli-fastboot/addon/services/fastboot.js
+++ b/packages/ember-cli-fastboot/addon/services/fastboot.js
@@ -69,6 +69,13 @@ const FastBootService = Service.extend({
   headers: deprecatingAlias('request.headers', { id: 'fastboot.headers-to-request', until: '0.9.9' }),
   isFastBoot: typeof FastBoot !== 'undefined',
 
+  isFastboot: computed(function() {
+    assert(
+      'The fastboot service does not have an `isFastboot` property. This is likely a typo. Please use `isFastBoot` instead.',
+      false
+    );
+  }),
+
   init() {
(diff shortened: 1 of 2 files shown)


packages/ember-cli-fastboot/addon/services/fastboot.js
```diff
@@ -69,6 +69,13 @@ const FastBootService = Service.extend({
   headers: deprecatingAlias('request.headers', { id: 'fastboot.headers-to-request', until: '0.9.9' }),
   isFastBoot: typeof FastBoot !== 'undefined',
 
+  isFastboot: computed(function() {
```
bertdeblock, line 72: Or should we warn here instead and return `this.isFastBoot`?

rwjblue: I think an assertion is fine, though I'd prefer if we didn't make this a computed (it will only assert the first time it is accessed I think).

We should consider migrating to native classes here, and then using a native getter directly. Unfortunately, we'll have to do this conditionally based on Ember version. We need to drop Ember < 3.12 support, would you mind making an issue for discussing that?

bertdeblock: Created https://github.com/ember-fastboot/ember-cli-fastboot/issues/815 for this. Should I just leave the PR until using a native getter is possible or?

mansona (approved):
Woop thanks for the contribution 🎉

rwjblue (approved):

PR #817, opened 2021-03-25
Replace term whitelist with allowlist

kiwiupover:
I followed the RFC from awhile back that has since been merged, so
hopefully this can be changed here in ember-cli-fastboot as well.


mcfiredrill:
Thank you so much! Sorry I must have missed your previous comment. 
https://github.com/ember-fastboot/ember-cli-fastboot/pull/806#issuecomment-769917539

Let me know if I can help with anything and feel free to close the previous PR #806

bertdeblock:
Is this in response to the [Replace terms blacklist & whitelist in Ember CLI](https://emberjs.github.io/rfcs/0639-replace-blacklist-whitelist.html) RFC? If so, I thought the settled upon terms were `include` and `exclude`? If not, please ignore me 😄.

mansona (changes requested):
First of all, thanks a million for doing all this work 🎉  this is amazing and really detailed!! 

I'm looking at this and I can see that it's clearly a breaking change 🤔  I'm wondering if it would be better to implement some sort of deprecation around `hostWhitelist` so  that we don't need to release this as a breaking change and we can give people a bit of time to update their system. What do you think?

xg-wang:
fastboot has already released v3 https://github.com/ember-fastboot/ember-cli-fastboot/milestone/2, as a minor bump we can have an alias for the old name (with deprecation would be even better); for ember-cli-fastboot it seems fine to include a breaking change, as the public interface in config is straight forward, e.g., `test-packages/basic-app/config/environment.js`

mansona:
@xg-wang I would feel quite strongly against releasing a breaking change for this in a major version that has never been deprecated once since that's not how we usually do things in Ember. I think it's totally fine for the alias and the deprecation to live for the entirety of the 3.x series 👍  especially since it won't be a massive code-path that needs to still hang around, as we can use the new name internally and just make sure that we are reading from the old name as a fallback

packages/fastboot/test/fastboot-request-test.js
```diff
@@ -2,7 +2,7 @@ var expect = require('chai').expect;
 var FastBootRequest = require('./../src/fastboot-request.js');
 
 describe('FastBootRequest', function() {
-  it('throws an exception if no hostWhitelist is provided', function() {
+  it('throws an exception if no hostAllowList is provided', function() {
```
snewcomer, line 5:
```suggestion
  it('throws an exception if no hostAllowlist is provided', function() {
```

kiwiupover:
@mansona I have fixed the requested changes for this PR. Can you please review again.

packages/fastboot/src/fastboot-schema.js
```diff
@@ -83,15 +83,15 @@ function loadConfig(distPath) {
   }
 
   let sandboxRequire = buildWhitelistedRequire(
-    pkg.fastboot.moduleWhitelist,
+    pkg.fastboot.moduleAllowlist,
```
snewcomer, line 86: I think we might want to add a fallback + deprecation here as well?

packages/ember-cli-fastboot/lib/broccoli/fastboot-config.js
```diff
@@ -160,32 +159,35 @@ module.exports = class FastBootConfig extends Plugin {
     this.manifest = this.updateFastBootManifest(manifest);
   }
 
-  buildHostWhitelist() {
+  buildHostAllowList() {
     if (this.fastbootAppConfig) {
-      this.hostWhitelist = this.fastbootAppConfig.hostWhitelist;
+      if ('hostWhitelist' in this.fastbootAppConfig) {
+        this.ui.writeLine('Please update your fastboot config to use `hostAllowList` of the deprecated `hostWhitelist`');
```
snewcomer, line 165: to use `hostAllowList` instead of the deprecated...

README.md
```diff
@@ -273,23 +273,23 @@ module.exports = function(environment) {
     },
 
     fastboot: {
-      hostWhitelist: ['example.com', 'subdomain.example.com', /^localhost:\d+$/]
+      hostAllowlist: ['example.com', 'subdomain.example.com', /^localhost:\d+$/]
```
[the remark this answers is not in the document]
snewcomer: I wouldn't be opposed to just making this `Allowlist` (lower case l) to ensure ease of migration

snewcomer (commented):
This looks like a great PR!  Are we concerned about making this breaking change while still in an alpha/beta 3.0? Especially with acceptable fallbacks and warnings to the end user?  

As long as we agree on the approach of the PR, my preference is to move this forward given this is well documented (e.g. `moduleAllowlist` perhaps could be added to the README; CHANGELOG would need to clearly call this out).

PR #818, opened 2021-03-26
Turn on rehydration based on config

mansona:
This is something that we discussed in the weekly Fastboot meeting where I thought it would be useful to be able to turn on rehydration based on a config entry rather than using the environment variable.

This would allow you to create a PR with the change to turn on rehydration quite easily, and also would support most preview apps setups when using heroku or something like netlify if you're using prember

I've almost finished this PR, I just can't quite get the config into the fastbootOptions 🤔


packages/ember-cli-fastboot/index.js
```diff
@@ -336,8 +336,13 @@ module.exports = {
             this.fastboot = new FastBoot(fastbootOptions);
           }
 
+          let appConfig = this._getHostAppConfig();
+
           let fastbootMiddleware = FastBootExpressMiddleware({
-            fastboot: this.fastboot
+            fastboot: this.fastboot,
+            visitOptions: {
+              renderMode: appConfig.fastboot.renderMode
```
xg-wang, line 344: Suggestion: check input and throw a silent-error if not allowed values (`"serialize"`)

packages/ember-cli-fastboot/test/fastboot-rendermode-config-test.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
```
xg-wang, line 1: We can start writing tests involving fastboot / ember-cli-fastboot in https://github.com/ember-fastboot/ember-cli-fastboot/blob/2a6ed5d76b00da2106a9ba10f048aa2bef989a80/test-packages/integration-tests/test/basic-test.js.
But we can land this PR and do another follow up to move later.

packages/ember-cli-fastboot/index.js
mansona: yes that's probably a good idea, but to be honest I'm not even sure if `renderMode` is the right thing to be passing around 🤔  right now it can only have exactly one value and that is being used to turn a feature on... which makes me feel like a boolean would be better

maybe something quite explicit like `turnOnRehydration` or something but I don't know 🤷  I guess the question becomes: how likely is it that we will have more render modes? 😂

packages/ember-cli-fastboot/test/fastboot-rendermode-config-test.js
mansona: oh is this the new place that @kiwiupover has been working on? to be honest I wasn't sure and I just copy and pasted an existing test to get it to work. I would much prefer to do it the right way in this PR rather than fix it up in another one, I just don't exactly know how 😂

mansona:
@xg-wang regarding our discussion of the API of this change, here is me consuming the fastboot lib in prember: https://github.com/ef4/prember/pull/66 which might be useful to see what the API looks like in the wild.

Let me know what you think, and we can probably discuss it in the weekly meeting too 👍

packages/ember-cli-fastboot/index.js
xg-wang: Yea I agree with you here `renderMode` is not the right public interface we expose.

`renderMode` is what ember-cli-fastboot passes to ember: when rehydration is enabled, it is `serialize` on server side, and `rehydrate` on client side. If rehydration is not enabled, ember will use `undefined` renderMode to use the default clientBuilder (see https://github.com/emberjs/ember.js/blob/4cb7da45fd6e4720daf7e769877b3d8868a3cf71/packages/%40ember/-internals/glimmer/lib/setup-registry.ts#L25-L38)

User should not be aware of these wirings, let's call it `enableRehydration` which people can set in `environement.fastboot.enableRehydration`.

The config will be passed to `new FastBoot({ ..., enableRehydration})`.

I'd like to deprecate the `EXPERIMENTAL_RENDER_MODE_SERIALIZE` with the PR to use the config flag instead.

For Prember, I'm not sure can you let user to turn on the flag then read from the same config?

xg-wang: Alternatively we can set it in `fastboot.visit(path, { enableRehydration })`, then per visit is allowed have different behavior. But I don't think it is necessary since user will have a single flag to control all visit behavior.

mansona: I already have a corresponding PR on Prember that I know works if you wanted to check it out https://github.com/ef4/prember/pull/66

I totally agree with your suggestion for the public API and I'll get that done when I get a chance 👍 

I hadn't actually thought about deprecating the environment variable 🤔 I guess it's a good idea so I'll add that to this PR too

While I agree that in my use case and in most use cases we would probably enable rehydration holistically for the whole app at once, @rwjblue did specifically ask that we expose a per-request API for people that didn't want to turn it on for everything or wanted to experiment with A/B tests for example. I have no preference on this particular point because I'm very much going to be enabling it for a full Prember run all at once so I can't really defend this position well since it's not my idea 🙈

packages/ember-cli-fastboot/test/fastboot-rendermode-config-test.js
xg-wang: You can move the fixture app to `test-packages`, then assert result in `test-packages/integration-tests/test/basic-test.js`. But don't worry I can get a test when we figure out how we want to name the config flag.

PR #821, opened 2021-04-06
Convert co to async in test; cleanup some configurations

xg-wang:


mansona:
ooo this is nice 😍  these sorts of PRs make me so happy 🎉

PR #822, opened 2021-04-07
Increment latest schema version to 5

xg-wang:
html-oriented manifest format
-----------------------------

In the new schema format, which is defined in
https://github.com/ember-fastboot/fastboot/pull/272/commits/3fd5bc90b63e2c497a8f519627556f9e94160065
the manifest is written into HTML and later extracted by fastboot on
server side instead of previously reading from dist/package.json

Note:
The new schema in fastboot does not handle fastboot config
https://github.com/ember-fastboot/ember-cli-fastboot/tree/e4d0b7c7bcdf82def0dc8726835b49d707673f41#providing-additional-config

This commit changes to read Fastboot.config from dist/package.json
instead of ignoring it

Allow to require module path from whitelisted dependency
-------------------------------------------------------

Incrementing schema to 5 also included the changes in schema 4
strictWhitelist
See https://github.com/ember-fastboot/fastboot/pull/200


packages/fastboot/test/fixtures/html-entrypoint/package.json
```diff
@@ -4,6 +4,30 @@
   "fastboot": {
     "schemaVersion": 5,
     "moduleWhitelist": [],
-    "htmlEntrypoint": "index.html"
+    "htmlEntrypoint": "index.html",
+    "config": {
```
xg-wang, line 8: @ef4 @rwjblue should we put config in a `<meta name="<my-app>/config/fastboot-environment">` instead? Then we can merge with browser env and pass it to sandbox

xg-wang:
We should also ignore certain scripts like https://github.com/embroider-build/embroider/pull/455/files

packages/fastboot/test/fixtures/html-entrypoint/package.json
ef4: I haven't thought about this in a while but that seems fine to me.

packages/fastboot/test/fastboot-test.js
```diff
@@ -455,18 +455,4 @@ describe('FastBoot', function() {
       usedPrebuiltSandbox: true,
     });
   });
-
-  it('htmlEntrypoint works', function() {
```
xg-wang, line 459: Moved to `test-packages/integration-tests/test/basic-test.js`

packages/fastboot/test/fixtures/html-entrypoint/package.json
xg-wang: Done in latest commit. v5 schema after the fix is backwards compatible to embroider build

xg-wang:
@ef4 @rwjblue this is ready for review, can you take a look?

xg-wang:
Start to resolver error. Not sure if it's because ember-cli-addon-tests always use latest packages.

```
 FastBootLocation
    1) should NOT redirect when no transition is called
(node:78372) UnhandledPromiseRejectionWarning: Error: RootResolveError: RootResolveError: unable to resolve root /private/var/folders/bg/jphvvr3s1g15h75hpk_sy468000sx4/T/d-2021720-78043-szhq0h.x116/under-test/fastboot-location/node_modules/jquery/dist: RootResolveError: "/private/var/folders/bg/jphvvr3s1g15h75hpk_sy468000sx4/T/d-2021720-78043-szhq0h.x116/under-test/fastboot-location/node_modules/jquery/dist" resolved to "/private/var/folders/bg/jphvvr3s1g15h75hpk_sy468000sx4/T/d-2021720-78043-szhq0h.x116/pristine/node_modules/jquery/dist" but we were unable to examine "/private/var/folders/bg/jphvvr3s1g15h75hpk_sy468000sx4/T/d-2021720-78043-szhq0h.x116/under-test/fastboot-location/node_modules/jquery/dist" using strict case sensitive rules.  Please check each component of the path and make sure that that path exactly matches the correct case of the files on your filesystem.
    at BunserBuf.<anonymous> (/private/var/folders/bg/jphvvr3s1g15h75hpk_sy468000sx4/T/d-2021720-78043-szhq0h.x116/pristine/node_modules/fb-watchman/index.js:95:23)
    at BunserBuf.emit (events.js:375:28)
    at BunserBuf.emit (domain.js:470:12)
    at BunserBuf.process (/private/var/folders/bg/jphvvr3s1g15h75hpk_sy468000sx4/T/d-2021720-78043-szhq0h.x116/pristine/node_modules/bser/index.js:292:10)
    at /private/var/folders/bg/jphvvr3s1g15h75hpk_sy468000sx4/T/d-2021720-78043-szhq0h.x116/pristine/node_modules/bser/index.js:247:12
    at processTicksAndRejections (internal/process/task_queues.js:77:11)
```

Issue #819, opened 2021-04-04
DEPRECATION: A value was injected implicitly on the '_fastbootInfo' property


lolmaus:
Hi!

I'm receiving this deprecation:

> DEPRECATION: A value was injected implicitly on the '_fastbootInfo' property of an instance of <docs-app@service:fastboot::ember223>. Implicit injection is now deprecated, please add an explicit injection for this value. If the injected value is a service, consider using the @service decorator. [deprecation id: implicit-injections] See https://deprecations.emberjs.com/v3.x#toc_implicit-injections for more details.

This appeared after upgrading Ember from 3.23 to 3.26.

I am not touching `_fastbootInfo` in my code and am not using any third-party addons on top of FastBoot.

mamiller93:
Any update on this one?

knownasilya:
Looks like https://github.com/ember-fastboot/ember-cli-fastboot/blob/master/packages/fastboot/src/fastboot-info.js#L41 is set here. Probably need to change how this value is accessed on the fastboot service. Not sure what the best option is. cc @kiwiupover

PR #823, opened 2021-04-07
Schema version 6: lookup fastboot info

xg-wang:
Implicit injection is deprecated https://deprecations.emberjs.com/v3.x#toc_implicit-injections

Fixes https://github.com/ember-fastboot/ember-cli-fastboot/issues/819
Depends on https://github.com/ember-fastboot/ember-cli-fastboot/pull/822

This is breaking change because user needs to upgrade to compatible `fastboot`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
       - name: Yarn Install
         run: yarn install --ignore-engines --frozen-lockfile
       - name: Integration Tests
-        run: yarn workspace integration-tests test 
+        run: yarn workspace integration-tests test
 
   test-packages:
     name: Test Packages
--- a/package.json
+++ b/package.json
@@ -9,7 +9,16 @@
     "packages/*",
     "test-packages/*"
   ],
+  "scripts": {
+    "test": "npm-run-all test:*",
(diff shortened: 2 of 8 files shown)


rwjblue:
I think we can do this without a breaking change. The deprecation only fires when you access an implicit injection, not when you call `instance.inject`. So we can always call `instance.inject` then we can update `fastboot` service like you did (manually clobber the injected value).

rwjblue:
See these tests in Ember for examples of how to avoid the deprecation on create:

https://github.com/snewcomer/ember.js/blob/bb607647f107a5046a19e67042bab33ba75f6e62/packages/%40ember/-internals/runtime/tests/system/object/create_test.js#L135-L158

https://github.com/snewcomer/ember.js/blob/bb607647f107a5046a19e67042bab33ba75f6e62/packages/%40ember/-internals/runtime/tests/system/object/create_test.js#L297-L351

These techniques can be used to significantly reduce the changes required to avoid the deprecaiton.

xg-wang:
@rwjblue oh that's nice, thanks for the examples. I'll update the PR

rwjblue:
Moved to draft for now (lemme know when ready for another round of review)

xg-wang:
Closed by https://github.com/ember-fastboot/ember-cli-fastboot/pull/841

PR #824, opened 2021-04-07
Update node engine requirement

xg-wang:
https://nodejs.org/en/about/releases/

Release | Status | Codename | Initial Release | Active LTS Start | Maintenance LTS Start | End-of-life
-- | -- | -- | -- | -- | -- | --
v10 | Maintenance LTS | Dubnium | 2018-04-24 | 2018-10-30 | 2020-05-19 | 2021-04-30
v12 | Maintenance LTS | Erbium | 2019-04-23 | 2019-10-21 | 2020-11-30 | 2022-04-30
v14 | Active LTS | Fermium | 2020-04-21 | 2020-10-27 | 2021-10-19 | 2023-04-30
v15 | Current |   | 2020-10-20 |   | 2021-04-01 | 2021-06-01
v16 | Pending |   | 2021-04-20 | 2021-10-26 | 2022-10-18 | 2024-04-30

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [14.x, 12.x, 10.x]
+        node-version: [16.x, 14.x, 12.x]
         os: [ubuntu-latest, windows-latest]
 
     steps:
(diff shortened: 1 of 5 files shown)


mansona:
So i did a little bit of investigating and it seems like there might be somehting wrong with Babel and the promise polyfil 🤔 

Screenshot 2021-05-31 at 17 32 14

That is an array that is being passed into `validateObject` (which fails if you don't pass in an object)

Interestingly I had a different error that I needed to workaround before I even go that far and it was caused by a file that has actually been removed from ember-cli in v3.17, so I wonder if the root of our issue is that we are in a very outdated ember-cli version 🤔

mansona:
this was included in https://github.com/ember-fastboot/ember-cli-fastboot/pull/834 👍

PR #825, opened 2021-04-07
Drop module unification support

xg-wang:
https://blog.emberjs.com/update-on-module-unification-and-octane/

Also run prettier, but the actual change only contains removing `const isModuleUnification = this._isModuleUnification();`


mansona (approved):
looks good to me 🎉

PR #834, opened 2021-05-31
Update using ember-cli-update and drop support for Node 10

mansona:
So this PR has been a bit of a saga 🙈 

Essentially I have been trying to get https://github.com/ember-fastboot/ember-cli-fastboot/pull/824 working (so that we can add node 16 to our test suite) and I have been having the world of trouble. 

To make a long story very short: we had a problem with a sub-sub-sub-dependency of ember-cli-babel@6, workerpool. The version that we had in our sub-sub-sub-dependencies was very very old and had a bug in it that broke Node 16 that was fixed in the latest version.

My first attempt to fix this was to try and update the version of workerpool that we used in our old ember-cli-fastboot and potentially release a patch version of that: https://github.com/babel/broccoli-babel-transpiler/pull/203 but @rwjblue rightly said that this caused odd semver concerns. 

in that conversation @stefanpenner said that if we could patch the old version of workerpool then we could release a patch of that old version, so I did 🎉 https://github.com/josdejong/workerpool/pull/309

Now that workerpool@2.3.4 has been released our problem with Node 16 that was failing the tests has gone away 🎉 

## Todo 

- [x] add issue to fix application wrapper bug in ember-cli-fasboot tests https://github.com/ember-fastboot/ember-cli-fastboot/issues/838


packages/ember-cli-fastboot/package.json
```diff
@@ -43,47 +49,61 @@
...
+    "qunit": "^2.14.1",
+    "qunit-dom": "^1.6.0",
     "release-it": "^12.0.1",
     "release-it-lerna-changelog": "^1.0.2",
     "request": "^2.88.0",
     "rsvp": "^4.8.3"
   },
   "engines": {
-    "node": "10.* || 12.* || >= 14.*"
+    "node": "10.* || >= 12"
```
snewcomer, line 99: Should we update this to node 12 and later since we are still pre 3.0?  OR would you want that as a separate PR?

snewcomer: Happy to contribute if so

mansona: So... I was thinking it would be better to do them in a different PR. This PR was to try and fix the odd behaviour in these tests: https://github.com/ember-fastboot/ember-cli-fastboot/pull/824

The more I think about it I think we need to update with ember-cli-update and change the node requirements in the same PR

snewcomer: Seems sane!

snewcomer (approved):
This looks like great work!

mansona:
Thanks for the note @snewcomer 😂 but I don't entirely agree that this looks like great work 😂 

The thing is I don't know that putting a resolution in our package.json will "fix" the issue, because anyone downstream will still have this issue. I briefly discussed this with @Turbo87 in discord the other day and we had some ideas on how to proceed: https://discord.com/channels/480462759797063690/485861149821239298/850047640816582767

I would love to know your thoughts, It's a shame I couldn't join the meeting yesterday because I would have loved to discuss this.

packages/ember-cli-fastboot/package.json
mansona: I ended up needing to pull in the engine requirement changes in this PR anyway 👍

packages/ember-cli-fastboot/tests/index.html
```diff
@@ -21,6 +21,13 @@
     {{content-for "body"}}
     {{content-for "test-body"}}
 
+    <div id="qunit"></div>
```
sumitd94, line 24: NIT: we can self enclose the tag here 

```javascript
<div id="quint" />
```

packages/ember-cli-fastboot/tests/index.html
```diff
@@ -21,6 +21,13 @@
...
+    <div id="qunit-fixture">
+      <div id="ember-testing-container">
+        <div id="ember-testing"></div>
```
sumitd94, line 27: NIT: RE

sumitd94 (approved):
Two NIT Comments. This looks great 🎉 🔥

xg-wang (approved):

PR #836, opened 2021-06-08
Restore the legacy tests

kiwiupover:
The tests were removed in this PR https://github.com/ember-fastboot/ember-cli-fastboot/pull/805

These [tests](https://github.com/ember-fastboot/ember-cli-fastboot/tree/master/packages/ember-cli-fastboot/test) have not been exercised this the above PR.
https://github.com/ember-fastboot/ember-cli-fastboot/tree/master/packages/ember-cli-fastboot/test

Adding the npm package `[co](https://www.npmjs.com/package/co)` which is used when running the legacy tests

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,5 @@ jobs:
       - name: Run Mocha Tests
         working-directory: ./packages/ember-cli-fastboot
         run: |
-          volta install npm@4
           npm --version
           yarn test:mocha
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
       - name: Yarn Install
         run: yarn install --ignore-engines --frozen-lockfile
       - name: Integration Tests
-        run: yarn workspace integration-tests test 
+        run: yarn workspace integration-tests test
 
   test-packages:
     name: Test Packages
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -90,3 +90,34 @@ jobs:
         run: yarn workspace basic-app test:mocha
       - name: Custom App
         run: yarn workspace custom-fastboot-app test:mocha
+
+  test-legacy-mocha:
+    name: Legacy Mocha Tests - ${{ matrix.node-version }}
```
snewcomer, line 95: I wonder what is legacy about these

snewcomer (approved):
Lovely!  Glad you found this.

As discussed, we will hopefully revisit the test harness setup for these tests after this PR.

mansona (approved):
looks good to me!

.github/workflows/ci.yml
mansona: I agree that we probably shouldn't call them legacy 🤔  but I wouldn't think it's a blocker

packages/ember-cli-fastboot/package.json
```diff
@@ -46,6 +46,7 @@
     "body-parser": "^1.18.3",
     "broccoli-asset-rev": "^3.0.0",
     "broccoli-test-helper": "^1.5.0",
+    "co": "4.6.0",
```
xg-wang, line 49: https://github.com/ember-fastboot/ember-cli-fastboot/blob/b6b17af9dfb308919b41b07c610397add3adcf90/packages/ember-cli-fastboot/test/new-package-json-test.js#L8
Looks like `co` isn't actually being used, we should be able to remove

.github/workflows/ci.yml
```diff
@@ -90,3 +90,34 @@ jobs:
...
+        uses: actions/checkout@v2
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: volta-cli/action@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      # Remove test-packages folder so that we don't leak node_modules between apps
+      - name: Remove test-packages
+        run: |
+          rm -rf test-packages
```
xg-wang, line 114: Curious why do we need to remove?

PR #840, opened 2021-07-21
Add a default state for metadata

suchitadoshi1987:
Currently ember addons that rely on certain metadata from fastboot service that depends on a transitive dependency that sets this metadata will run into issues during local development against a dummy app.
This PR adds the default value of the metadata so that the invocation of the meta doesnt hard fail.


kratiahuja:
Should we add a test for the default value ?

kratiahuja:
Also I can merge once tests are added but would be good to run this by @rwjblue

suchitadoshi1987:
cc @rwjblue

suchitadoshi1987:
> Should we add a test for the default value ? This way in future if we accidentally change this, then we have a test to back this up.

yup. added tests to ensure ^^

kratiahuja:
LGTM! thanks for adding the tests

xg-wang:
@suchitadoshi1987 I think we need to avoid creating new test fixtures here. I'll DM you to figure out the unit test or use a proper test app

xg-wang (approved):

xg-wang:
@suchitadoshi1987 looks like this change broke existing test https://github.com/ember-fastboot/ember-cli-fastboot/pull/843/checks?check_run_id=3259458760

PR #841, opened 2021-08-02
Avoid implicit injection deprecation

snewcomer:
As noted in reference PR, the deprecation is issued when the FastBoot service is created.  However, if we also register a property with the same name as the implicit injection (`app.inject('service:foo', 'store', 'some:object')`), the deprecation goes away.

close https://github.com/ember-fastboot/ember-cli-fastboot/issues/819
ref https://github.com/ember-fastboot/ember-cli-fastboot/pull/823/files
rfc https://github.com/emberjs/rfcs/pull/680


packages/ember-cli-fastboot/addon/services/fastboot.js
```diff
@@ -88,6 +89,11 @@ const FastBootService = Service.extend({
     return RequestObject.create({ request: get(this, '_fastbootInfo.request') });
   }),
 
+  _fastbootInfo: computed(function() {
```
xg-wang, line 92: Do we need the test changes in https://github.com/ember-fastboot/ember-cli-fastboot/pull/823/files? Specifically registering `info:-fastboot` in the container for later lookup.

snewcomer: It doesn't look like it based on the tests!  Do you see other value here?

ef4:
This is not really fixed. Adding ember-cli-fastboot 3.2.0-beta.3 to a new ember 3.27 app still results in the implicit injection deprecation message the first time you render in fastboot.

I think code like this needs to go away: https://github.com/ember-fastboot/ember-cli-fastboot/blob/8be7162fbf4a41ae48ad8c0688a8d55cbabd8579/packages/fastboot/src/fastboot-info.js#L41

xg-wang:
@ef4 would a new schema version avoid breaking change? Essentially the attempt at https://github.com/ember-fastboot/ember-cli-fastboot/pull/823/files

PR #849, opened 2021-08-20
Fix use of Ember global, breaking Ember 4

### LINKED ISSUE #827: Use of Ember global triggers deprecation / fails for Ember 4
Direct access of the Ember global [here](https://github.com/ember-fastboot/ember-cli-fastboot/blob/2a6ed5d76b00da2106a9ba10f048aa2bef989a80/packages/ember-cli-fastboot/vendor/experimental-render-mode-rehydrate.js#L7-L10) triggers a deprecation in latest Ember versions (3.27+)

simonihmig:
I think it's not ideal to rely on assumptions about the global `require()` and AMD-style modules, especially in a brave new Embroider world. But after all this is what the [RFC706 suggests in the drawbacks section](https://emberjs.github.io/rfcs/0706-deprecate-ember-global.html#drawbacks), and should be good enough to fix this for the short term...

Fixes #827

### DIFF
--- a/packages/ember-cli-fastboot/vendor/experimental-render-mode-rehydrate.js
+++ b/packages/ember-cli-fastboot/vendor/experimental-render-mode-rehydrate.js
@@ -1,7 +1,7 @@
 (function() {
   if (typeof FastBoot === 'undefined') {
     var current = document.getElementById('fastboot-body-start');
-    var Ember = require('ember').default;
+    var Ember = require.has('ember') ? require('ember').default : Ember;
 
     if (
       current &&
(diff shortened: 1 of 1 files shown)


simonihmig:
Somehow, the first attempt (first commit) of using just `require('ember').default` didn't work in Ember 3.26 (CI was failing), as that threw an error that the module is not defined (`require.entries` was an empty object), but it *did* work for 3.27 and 4.0 (tested locally)! 🤔

The fallback check in the 2nd commit fixes this. But given we don't test across Ember versions in CI yet, the confidence in this change is somewhat lower than I would want. Maybe we have to wait for #848 first?

snewcomer (approved):
👍 💯.  This will fix it!

snewcomer:
@simonihmig Still investigating but seeing `ReferenceError: Ember is not defined at ` when adding your work to this PR.  

https://github.com/ember-fastboot/ember-cli-fastboot/pull/848/checks?check_run_id=3400612883

https://github.com/ember-fastboot/ember-cli-fastboot/pull/848/files#diff-6332b424009db608242427547830fa767850f2b49ae9943dc54c8d43ba0e349bR4

rwjblue (approved):

PR #853, opened 2021-08-22
Fix fastboot-script assets with cutom root url

xg-wang:
When using custom rootURL specified in environment.js, the
fastboot-script is still having relative path to local dist path, not
containing rootURL path. The fastboot-script should not be ignored.

Extracted from #822


snewcomer (approved):
Nice PR!

Issue #815, opened 2021-03-16
Update "fastboot" service to use native classes


bertdeblock:
[More info.](https://github.com/ember-fastboot/ember-cli-fastboot/pull/814/files#r595220575)

/cc @rwjblue

PR #856, opened 2021-09-10
[cleanup]: class based fastboot service

snewcomer:
This PR just seeks to cleanup some code and deprecations

close https://github.com/ember-fastboot/ember-cli-fastboot/issues/815

### DIFF
--- a/packages/ember-cli-fastboot/addon/services/fastboot.js
+++ b/packages/ember-cli-fastboot/addon/services/fastboot.js
@@ -29,7 +29,7 @@ const RequestObject = EObject.extend({
   })
 });
 
-const Shoebox = EObject.extend({
+class Shoebox extends EObject {
   put(key, value) {
     assert('shoebox.put is only invoked from the FastBoot rendered application', this.get('fastboot.isFastBoot'));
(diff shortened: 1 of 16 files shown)


ef4:
👍 

I'm not sure what the support matrix is between `fastboot` and `ember-cli-fastboot` but this change means `fastboot` will be dropping support for `ember-cli-fastboot` prior to #841.

snewcomer:
Actually there is one case we could go about it another way.  When I originally did the PR, this was one of the test cases.  We need to add a setter for it to not raise a deprecation.

https://github.com/emberjs/ember.js/pull/19358/files#diff-2600f49bdcd760f57af70712f38930aca2c85c30752165032d9998359ed9109dR265-R295

snewcomer:
@ef4 I've changed the scope of this PR and opened a new PR with the actual fix that works - https://github.com/ember-fastboot/ember-cli-fastboot/pull/857.  

> change means fastboot will be dropping support for ember-cli-fastboot

However, all `inject`ions do need to be dropped for 4.0.  From your point of view, what is the concern if we aren't injecting directly onto the fast boot service from ember-cli-fastboot?

PR #857, opened 2021-09-13
Avoid implicit injection _fastbootInfo deprecation pt 2

snewcomer:
This actually fixes #841.

Note: this way this works is if there is a setter (not just a getter), then it overrides and doesn't issue the deprecation.

https://github.com/emberjs/ember.js/pull/19358/files#diff-2600f49bdcd760f57af70712f38930aca2c85c30752165032d9998359ed9109dR265-R295


mansona (approved):

SergeAstapov:
Looks like it's still an issue - #869

PR #858, opened 2021-09-14
Add ember-try testing for ember-cli-fastboot

mansona:
This is also adding some "no-deprecations" tests so that hopefully we can see that #857 will actually fix things objectively 😂


NullVoxPopuli:
anything keeping this from being merged? it'd be a requirement  for further updating for ember 4 support (require ember-auto-import@v2)

PR #862, opened 2021-10-20
Ensure rehydration works on Ember < 3.27

rwjblue:
The fallback for the guard introduced in 32d6f074c66aed32898a05934dad24b4b8cb19e4 (which takes effect when running on Ember < 3.27) is somewhat fundamentally borked. Due to the usage of `var Ember` and the fact that `var` hoists, the fallback in the ternary **can never** anything but `undefined`.


rahulk94:
@rwjblue I just ran into this whilst trying to use Fastboot beta 4 with a Ember 3.24.x app. Can you look at doing a beta release sooner than later?

PR #865, opened 2021-11-26
Update .npmignore to ignore .github and test folders

SergeAstapov:
This shrinks the "unpacked size" for a few Kb


mansona (approved):
So usually I would say that this sort of change isn't really that important since a) the "unpacked size" isn't all that representative of reality most of the time in the ember world and b) it matters less with something that is more of a server technology.

In this case there is literally no downside so I'm going to merge it 👍

SergeAstapov:
@mansona totally agree with you! This is just to keep things tidy

PR #866, opened 2021-11-27
Make edition=octane and update deps for Ember v4 compatibility

SergeAstapov:
This makes CI green again.


packages/ember-cli-fastboot/test/fastboot-config-test.js
```diff
@@ -17,7 +17,7 @@ describe('FastBoot config', function () {
     return app
       .create('fastboot-config', {
         skipNpm: true,
-        emberVersion: 'latest',
```
mansona, line 20: soo... while this may be making CI green again it's not actually getting to the root of the problem 🤔 the issue is that pinning the ember version to pre-4.0 it will pass but that doesn't accurately test the underlying issue that was causing it to fail with latest

I hope what I'm saying makes sense 🙈

mansona (changes requested):
I don't know what the changes need to be to make this ready to merge but we need to think just a tiny bit more deeply about it and make sure we're doing the right thing. I tried to explain myself in a comment but I fear that I haven't been clear enough and this is a bit of a nuanced problem 🙃

packages/ember-cli-fastboot/test/fastboot-config-test.js
SergeAstapov: @mansona I'm looking into this and will get back once have some findings

SergeAstapov: @mansona here my findings:
* in order to keep `latest` here we should add `ember-auto-import` v2 and `webpack` dependencies in `editPackageJSON` step
* Once this is done, build fails with `Cannot read property 'length' of undefined` error caused by [this line](https://github.com/ef4/ember-auto-import/blob/v2.3.0/packages/ember-auto-import/ts/package.ts#L407)

Unfortunately I can't seem to find a way to fix missing `rootURL` hence not sure I can't make it work with Ember v4 :(

Came across https://github.com/tomdale/ember-cli-addon-tests/issues/220 and I wonder if it makes sense to completely refactor away test harness from using `ember-cli-addon-tests` either to `scenario-tester` or even to just test projects (which we already have in `test-packages`).

Seem the latter option may be simpler as new contributors do not need to learn new tool `scenario-tester` but may be harder to support (as we would need to update those test apps).

rwjblue: >  I wonder if it makes sense to completely refactor away test harness from using ember-cli-addon-tests either to scenario-tester or even to just test projects (which we already have in test-packages).

Yes, agreed. I'm fine with either of those paths forward, but I think individual test-packages will be a smaller migration for the immediate problem. In the long run, I'd like to end up with scenario-tester (but again that would be a pretty significant departure and can be done independently from unblocking v4 support/testing stuff).

SergeAstapov: @rwjblue would you feel confident to land this PR "as is" to make CI green again or you think we should get these tests pass with Ember v4.x?

SergeAstapov:
@mansona CI now green with latest Ember.js version! Commits history may point to all the problems I faced.

mansona (approved):
omg fantastic work 🎉 that is a really nice find, I don't think I would have noticed that for weeks 🙈

PR #867, opened 2021-11-27
Remove unused nested github actions setup

SergeAstapov:
I guess these are just leftovers from migration to monorepo


mansona:
@xg-wang I know you're not around quite as much but can you just confirm that these are just hangovers from the monorepo stuff? 

I tend to agree with @SergeAstapov since their location makes them inert but I wanted to check with you first 👍

mansona (approved):
I think this is probably fine to merge, if we need them we can just create them again 😂  #githistoryforlife

PR #868, opened 2021-12-19
Chore: Remove warnings from 7 years ago

steveclifton:
When I first read these warnings - I thought it wasnt safe to use in production.

Then I saw the blame, and it was 7 years ago.

Consider updating the readme? :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,6 @@ FastBoot allows you to render and serve Ember.js apps on the server.
 Using FastBoot, you can serve rendered HTML to browsers and
 other clients without requiring them to download JavaScript assets.
 
-Currently, the set of Ember applications supported is extremely limited.
-As we fix more issues, we expect that set to grow rapidly. See [Known
-Limitations](#known-limitations) below for a full-list.
-
-The bottom line is that you should not (yet) expect to install this add-on in
-your production app and have FastBoot work.
-
 ## Installation
 
 FastBoot requires Ember 2.3 or higher. It is also preferable that your app is running `ember-cli` 2.12.0 and higher.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -624,12 +617,6 @@ fastbootConfigTree() {
 
 The above configuration will be available in Node via the `FastBoot.config()` function. Therefore, in order to get the above config, the addon/app can call `FastBoot.config('<engine-name>')`.
 
-## Known Limitations
```
SergeAstapov, line 627: @steveclifton Maybe my personal preference, but I would keep this header as otherwise below section will be part of "Build Hooks for FastBoot" which not exactly what it is.
What I mean, we can keep the header "## Known Limitations" and only get rid of the statement below it

mansona: I would say that we can just remove the sentence `Only the most brave should even
consider deploying this to production.` and that works. And I agree with @SergeAstapov that we should keep the section

README.md
```diff
@@ -7,13 +7,6 @@ FastBoot allows you to render and serve Ember.js apps on the server.
 Using FastBoot, you can serve rendered HTML to browsers and
 other clients without requiring them to download JavaScript assets.
 
-Currently, the set of Ember applications supported is extremely limited.
-As we fix more issues, we expect that set to grow rapidly. See [Known
-Limitations](#known-limitations) below for a full-list.
-
-The bottom line is that you should not (yet) expect to install this add-on in
```
mansona, line 14: This sentence is actually still true, or at least how I am reading it: "you can't expect to install this and have everything 'just work' and you will likely need to tweak some things"

Consider rephrasing rather than removing 👍

mansona (changes requested):
overall I think the point of this PR is great, but it would need to improve slightly before we merge it. I'm happy to remove any indication that fastboot is not ready for prime time for sure.

Also can you update your PR name so that it's more representative of the actual change, something like "update readme to remove warnings about running fastboot in production"

rahulk94:
@mansona @SergeAstapov I've raised https://github.com/ember-fastboot/ember-cli-fastboot/pull/891 to update this info based on your feedback

rahulk94:
#891 has been merged. This can now be closed @steveclifton.

PR #886, opened 2022-01-24
Update and synchronize dependencies in the monorepo

SergeAstapov:


packages/fastboot/package.json
```diff
@@ -28,30 +28,30 @@
...
   "dependencies": {
-    "chalk": "^3.0.0",
-    "cookie": "^0.4.0",
-    "debug": "^4.1.1",
-    "jsdom": "^16.2.2",
-    "resolve": "^1.15.0",
+    "chalk": "^4.1.2",
+    "cookie": "^0.4.1",
+    "debug": "^4.3.3",
+    "jsdom": "^19.0.0",
```
mansona, line 34: I'm trying to be quite careful about the dependency versions and it looks like this one is the only one that could potentially break something. It's a funny one because the release that we're aiming to release is 3.2 for all the versions which will only be a minor version for fastboot but a major version for every other package. We need to be sure that this is in fact not breaking for fastboot if we're still going to have that strategy, otherwise we need to investigate a plan we had to be able to specify different version numbers on a release

mansona (changes requested):
I'm just "requesting changes" on this because we probably need a discussion before it's ok to ship. And it's only one dependency on one project that I'm curious about too

SergeAstapov:
Discussed with @mansona and bumping `jsdom` should not be a concern as we use it only to parse an html for `config/environment` and manifest

PR #892, opened 2022-03-29
Pass the log argument to worker in fastboot-app-server

Aierie:
This is necessary for fastboot middleware to receive logging options specified in fastboot-app-server


SergeAstapov (approved):

SergeAstapov:
this is follow up to #890

PR #894, opened 2022-06-09
fix: properly attribute stack with fileName, removes slow leak

gabrielcsapo:
Co-authored-by: Brenden Palmer <brendenpalmer@gmail.com>

:::DO NOT USE:::

SEEMS LIKE SOURCE MAPS AND VM IS NOT WORKING PROPERLY

# Summary

We found a slow leak in fastboot, upon further analysis of the heap we found that the https://github.com/evanw/node-source-map-support in https://github.com/evanw/node-source-map-support/blob/ac2c3e4c633c66931981ac94b44e6963addbe3f4/source-map-support.js#L37-L41 will increase over time by design as errors are created. If we utilize the built in source-map option in node we get better results and node will handle this weak map internally.

Screen Shot 2022-06-09 at 10 52 11 AM

## Before

> stack with incorrect path on disk

Screen Shot 2022-06-09 at 10 41 42 AM

Screen Shot 2022-06-09 at 10 42 43 AM

## After

> stack with correct path on disk

Screen Shot 2022-06-09 at 10 42 00 AM
Screen Shot 2022-06-09 at 10 42 09 AM
Screen Shot 2022-06-09 at 10 42 11 AM


krisselden (approved):
Thank you, nice detective work!

rwjblue (changes requested):
I think as written, this is a breaking change right (folks that used to get source mapped errors won't now)? Can we make the change "opt-in" somehow?

Otherwise, we'll have to do a major bump which may be harder to get the change out to folks...

rwjblue:
I was chatting with @brendenpalmer and @gabrielcsapo and I think we might be able to do some sort of fall back logic like:

1. detect if node >= 14 -- if so -> just call `process.setSourcEMapsEnabled(true)`
2. if node ~ 12 -> check if CLI flag or `NODE_OPTIONS` is setup
3. else -> do the old brokenish thing

packages/fastboot/src/sandbox.js
```diff
@@ -17,9 +17,11 @@ module.exports = class Sandbox {
     let URL = require('url');
     let globals = this.globals;
 
+    const sourceMapConfig = process.setSourceMapsEnabled ? {} : { sourceMapSupport };
```
rwjblue, line 20: This has a few issues:

* when we are using `process.setSourceMapsEnabled` we should not expose any globals
* we should not be changing the name of the global to export (when using Node < 14) see (which uses the old global for exposing `source-map-support`):

https://github.com/ember-fastboot/ember-cli-fastboot/blob/f94426759fb89d94db71b93f97cae92fb183e114/packages/fastboot/src/scripts/install-source-map-support.js#L2-L6


My suggestion is (basically):

```js
    const sourceMapConfig = process.setSourceMapsEnabled ? null : { sourceMapSupport };

    let sandbox = Object.assign(
      sourceMapConfig,
      {
        console,
        setTimeout,
        clearTimeout,
        URL,
        // Convince jQuery not to assume it's in a browser
        module: { exports: {} },
      },
      globals
    );
```

This will:

1. do nothing if we are going to use `process.setSourceMapsEnabled`
2. when we are using the older system, we still expose the `source-map-support` package at the correct location

rwjblue (changes requested):

rwjblue (changes requested):

PR #895, opened 2022-06-13
fix: removes slow leak by emptying cache for source maps

gabrielcsapo:
A faster fix for the slow memory leak explained in https://github.com/ember-fastboot/ember-cli-fastboot/pull/894, but without depending on the vm to have source maps knowledge which is currently a bug.

### DIFF
--- a/packages/fastboot/src/scripts/install-source-map-support.js
+++ b/packages/fastboot/src/scripts/install-source-map-support.js
@@ -1,12 +1,10 @@
 'use strict';
 /* globals sourceMapSupport */
 
-Error.prepareStackTrace = function prepareStackTrace(error, stack) {
-  return error + stack.map(frame => '\n    at ' + sourceMapSupport.wrapCallSite(frame)).join('');
-};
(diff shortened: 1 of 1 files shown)


rwjblue (approved):

packages/fastboot/src/scripts/install-source-map-support.js
```diff
@@ -1,12 +1,10 @@
 'use strict';
 /* globals sourceMapSupport */
 
-Error.prepareStackTrace = function prepareStackTrace(error, stack) {
-  return error + stack.map(frame => '\n    at ' + sourceMapSupport.wrapCallSite(frame)).join('');
-};
```
rwjblue, line 6: Note: this is removed because `source-map-support` never uses the pre-existing `Error.prepareStackTrace` value ([it always clobbers](https://github.com/evanw/node-source-map-support/blob/ac2c3e4c633c66931981ac94b44e6963addbe3f4/source-map-support.js#L581-L585)).

PR #902, opened 2022-12-01
Update CI and drop support for Node 12

mikrostew:
# Summary

Fix CI for this repo, which has been failing for the last few months, and run against Node 14.x and 16.x.

## Updated Node LTS Versions

This drops support for Node 12.x (which is EOL as of 2022/04/30), and runs the tests against 14.x and 16.x.

I also pinned a more recent version of Yarn.

_Notes_
I had to update `workerpool` in the lockfile to pull in the fix for Node16 mentioned in [this issue](https://github.com/babel/broccoli-babel-transpiler/pull/203).

## Legacy Mocha Tests

These have been failing for a while, with this error:

```
Error: Unable to require module 'crypto' because it was not in the whitelist.
```

So I added that to `fastbootDependencies`, which fixed the issue.

_Notes_
Running with Node 16 surfaced a new error with starting the server in the tests, so I removed the `skipNpm: true` option to fix that.

## Ember-Try Tests

These have also been failing for a while, with this error:

```
Build Error (broccoli-persistent-filter:Babel > [Babel: @ember/test-helpers]) in @ember/test-helpers/-internal/build-registry.js

[BABEL]: Cannot find module 'babel-plugin-dynamic-import-node/utils'
```

I fixed that by preventing ember-try from running with `--no-lockfile` (which it does by default, and was causing a dependency mismatch when running the tests).


mikrostew:
I don't have permissions to run CI on this repo, but I was able to successfully run these changes against my fork: https://github.com/mikrostew/ember-cli-fastboot/pull/3

ef4:
Thanks, this looks OK so I'm going to merge it so it will actually get CI running again and you can see if it works.

PR #903, opened 2022-12-07
fix: Provide a more meaningful error when whitelist entry is missing

runspired:
In emberjs/data#8101 it was noted that the error message provided here did not give enough context for users to understand the issue was due to the Fastboot environment and the existence of the fastbootDependencies allow list. This should help greatly with that.

Ref: https://github.com/emberjs/data/issues/8101#issuecomment-1341675294


runspired:
cc @elwayman02

elwayman02:
@ef4 thanks for fixing the prettier formatting. It looks like all the tests are passing now - if this change is desirable as-written, would this be good to merge?

PR #900, opened 2022-08-18
Remove deprecated injections in Ember 4.x

backspace:
This is an attempt to fix #869. I don’t have much experience with addon development, but this seemed like a conventional way to check whether `.inject` could still be used.

Thanks for the years of work on this, all!

Issue #869, opened 2021-12-28
Deprecations with ember-4.0.0 


fpauser:
Using the current ember-cli-fastboot beta with ember stable (4.0.1) I get the following deprecation warnings:

```
DEBUG: -------------------------------
DEBUG: Ember      : 4.0.1
DEBUG: Ember Data : 4.0.2
DEBUG: -------------------------------
DEPRECATION: As of Ember 4.0.0, owner.inject no longer injects values into resolved instances, and calling the method has been deprecated. Since this method no longer does anything, it is fully safe to remove this injection. As an alternative to this API, you can refactor to explicitly inject `_ajaxRequest` on `adapter`, or look it up directly using the `getOwner` API. [deprecation id: remove-owner-inject] See https://deprecations.emberjs.com/v4.x#toc_implicit-injections for more details.
        at logDeprecationStackTrace (/tmp/broccoli-235627kdVdLsbbU54p/out-303-broccoli_merge_trees/assets/@ember/debug/lib/deprecate.js:108:1)
        at HANDLERS.<computed> (/tmp/broccoli-235627kdVdLsbbU54p/out-303-broccoli_merge_trees/assets/@ember/debug/lib/handlers.js:26:1)
        at raiseOnDeprecation (/tmp/broccoli-235627kdVdLsbbU54p/out-303-broccoli_merge_trees/assets/@ember/debug/lib/deprecate.js:135:1)
        at HANDLERS.<computed> (/tmp/broccoli-235627kdVdLsbbU54p/out-303-broccoli_merge_trees/assets/@ember/debug/lib/handlers.js:26:1)
        at invoke (/tmp/broccoli-235627kdVdLsbbU54p/out-303-broccoli_merge_trees/assets/@ember/debug/lib/handlers.js:38:1)
        at deprecate (/tmp/broccoli-235627kdVdLsbbU54p/out-303-broccoli_merge_trees/assets/@ember/debug/lib/deprecate.js:181:1)
        at Registry.injection (/tmp/broccoli-235627kdVdLsbbU54p/out-303-broccoli_merge_trees/assets/@ember/-internals/container/index.js:830:1)
        at App.inject (/tmp/broccoli-235627kdVdLsbbU54p/out-303-broccoli_merge_trees/assets/@ember/-internals/runtime/lib/mixins/registry_proxy.js:222:1)
[...]

PR #904, opened 2022-12-08
Remove deprecated implicit-injections

mikrostew:
Same idea as #900, but also tests against 4.4.

This should fix the Ember 4 deprecations in #869.


ef4:
I'm going to assume this is a breaking change since anybody who was relying on the implicit injections will need to do their own explicit injections.

(Which is fine because I'm also assuming that the previous PR dropping old node versions is a breaking change.)

nlfurniss:
@ef4 yes the last two PRs will need to be released as a new major

PR #905, opened 2023-01-09
remove deprecated implicit injection

amiarSlimane:
Already solved here :

```
// ember-cli-fastboot/addon/services/fastboot.js
// this getter/setter pair is to avoid deprecation from [RFC - 680](https://github.com/emberjs/rfcs/pull/680)
  _fastbootInfo: computed({
    get() {
      if (this.__fastbootInfo) {
        return this.__fastbootInfo;
      }

      return getOwner(this).lookup('info:-fastboot');
    },
    set(_key, value) {
      this.__fastbootInfo = value;
      return value;
    }
  }),
```
 but didn't remove the implicit injection:
```
// ember-cli-fastboot/packages/fastboot/src/fastboot-info.js
  instance.inject('service:fastboot', '_fastbootInfo', 'info:-fastboot');
```

PR #908, opened 2023-02-09
Replace implicit injection with a direct lookup

chriskrycho:
This resolves the deprecation introduced in Ember v3.27, replacing the implicit injection with using a "normal" `Owner.lookup` in the service.


chriskrycho:
Whoops, this is a duplicate of #905; let's land that instead!

PR #911, opened 2023-03-01
Update readme examples to use native class syntax

SergeAstapov:


README.md
```diff
@@ -174,24 +174,24 @@ rendering the entire page. Note how the call should be wrapped in a `fastboot.is
 check since the method will throw an exception outside of that context:
 
 ```js
-import Ember from 'ember';
+import Component from '@ember/component';
```
gilest, line 177:
```suggestion
import Component from '@glimmer/component';
```

gilest (approved):
Nice work. One tiny suggestion but looks good

PR #914, opened 2023-04-03
Drop implicit injection

ef4:
This is removed at Ember 5, and it wasn't even doing anything because the consuming side of this was already refactored to use owner.lookup instead.


nlfurniss:
See https://github.com/ember-fastboot/ember-cli-fastboot/pull/905

PR #918, opened 2023-05-26
add fail-fast: false to all CI sections

mansona:
I'm just trying to make sure that #917 hasn't added more failure to the CI so I want to check each of the jobs and how many tests are failing 👍


mansona:
we can close this because it was included in #917

PR #919, opened 2023-05-29
Move some failing tests from pre-built fixtures to scenario-tester

mansona:
The CI has been very broken lately 🙈 and most of these errors have been exceedingly hard to understand because a lot of our tests rely on ages-old fixtures that are essentially pre-built ember apps. There is a better way to get this sort of "fixture" test and we've been using it in embroider and ember-auto-import for a very long time, it's [scenario-tester](https://github.com/embroider-build/scenario-tester)

This PR sets up the scenario-tester infrastructure and moves some of the failing tests over to it. Eventually we would want to move all the fixture tests over but I'm doing the absolute minimum to get CI passing right now 👍


gilest:
Very excited for this. Anything community members can do to help it along?

PR #920, opened 2023-06-01
Add Linting to CI

mansona:
It was getting a bit annoying to come across a file that was 100% red underlines so I figure it was time to add linting to the CI.

This is a minimal change for now and we can improve as we go along 👍 

cc @kategengler


kategengler (approved):
LGTM though I would just run prettier fix and get it over with

PR #927, opened 2023-10-16
Upgrade base

Gaurav0:


packages/ember-cli-fastboot/test/fixtures/customized-outputpaths/config/targets.js
```diff
@@ -9,5 +9,4 @@ const browsers = [
 
 module.exports = {
   browsers,
-  node: 'current',
```
SergeAstapov, line 12: this is unsafe change as we indeed want to test Node.js version specified in CI config

PR #916, opened 2023-05-13
Fix ci and get working with Ember 5

Gaurav0:
Merges https://github.com/ember-fastboot/ember-cli-fastboot/pull/898 , https://github.com/ember-fastboot/ember-cli-fastboot/pull/907, and https://github.com/ember-fastboot/ember-cli-fastboot/pull/915

fastboot-shoebox-test was broken by https://github.com/ember-fastboot/ember-cli-fastboot/pull/905

Only one test is failing, and it is due to this bug: https://github.com/emberjs/ember.js/issues/20377

Update: the test is "Fixed", but only because the current release is now Ember 5.0

PR #928, opened 2023-10-17
Update base and fix ci

Gaurav0:
This is an update of #916

Unlike that PR, the scope is reduced to:

1. updating to ember-cli 3.28 and related dependency bumps
2. Fixing ci

It does not:

1. Fix the issue with fetch globals
2. Fix deprecations
3. Make it compatible with Ember 5

These will be addressed in separate PRs so that this one is easier to review.

Remaining failing tests:
- ember-try ember-release
- ember-try ember-beta
- ember-try ember-canary
- ember-try ember-release-no-deprecations

All other tests are passing.

Also, I've taken care to make smaller commits and to clean them up. Hopefully it is easy to review commit by commit.


package.json
```diff
@@ -54,8 +57,14 @@
...
+  "pnpm": {
+    "peerDependencyRules": {
+      "ignoreMissing": ["@babel/*", "@eslint/*"]
+    }
+  },
   "volta": {
     "node": "16.18.1",
-    "yarn": "1.22.19"
+    "yarn": "1.22.19",
+    "pnpm": "7.33.5"
```
NullVoxPopuli, line 68: Why an old version of pnpm?

NullVoxPopuli (commented):
Can the switch to/from pnpm be separated from blueprintesque changes?

Seems like there are a few things happening in here, and it'd be better for the changelog to have them separated, not just for the changelog tho, but managing the realm of potential debug-points during a git bisect.

Thanks!

package.json
Gaurav0: That's the version currently in ci. I figured updating it should be separate.

Gaurav0:
@NullVoxPopuli Maybe some of the fixes might be independently added, but one of the primary problems in ci right now is using old versions of ember-source.

PR #930, opened 2023-11-14
update pnpm to v8

mansona:


gilest:
Dropping node versions would be a **good thing** IMO. We need to add Node 20, too as it's now the active LTS.

mansona:
@gilest yes I totally agree, but I'm not really comfortable doing changes like that without CI being green first 🙈 

I'm currently in a process of tracking down the failing tests and moving them over to scenario-tester to get them passing again https://github.com/ember-fastboot/ember-cli-fastboot/pull/919 (ah I see you've seen that PR already because of your comment 😂)

mansona:
we're using 10 now 👍

PR #935, opened 2024-04-30
update to Ember 5.8 using ember-cli-update

mansona:


packages/ember-cli-fastboot/package.json
```diff
@@ -49,57 +49,57 @@
...
+    "stylelint": "^15.11.0",
+    "stylelint-config-standard": "^34.0.0",
+    "stylelint-prettier": "^4.1.0",
+    "webpack": "^5.91.0"
   },
   "peerDependencies": {
     "ember-source": "^4.0.0"
   },
   "engines": {
-    "node": "14.* || 16.* || >= 18"
```
SergeAstapov, line 102: just FYI looks like it would be a breaking change

mansona: thanks! that wasn't intentional, so I reverted it 👍

PR #946, opened 2024-05-21
Update url for fastboot-app-server in package.json for npm

SergeAstapov:
npm still points to https://github.com/ember-fastboot/fastboot-app-server


mansona (approved):
Looks good 👍

PR #948, opened 2024-05-22
Fix workspace protocol for fastboot-express-middleware package

SergeAstapov:
@mansona I guess this is the reason why `fastboot-express-middleware` was not included in latest release https://github.com/ember-fastboot/ember-cli-fastboot/releases/tag/v4.1.5-ember-cli-fastboot


mansona:
so to be honest I think there was too much stuff included in the last release 🤔 I didn't expect anything other than `fastboot-app-server` to be released since it's downstream from fastboot and ember-cli-fastboot

i didn't investigate since it was such a small issue I didn't think it mattered

SergeAstapov:
@mansona as far as I see, it publishes everything because `release-plan` pin dependencies to specific version (or maybe it's pnpm) when the package or it's dependencies within monorepo gets bumped/published.

because of the typo fixed in this PR, `fastboot-express-middleware` was not bumped and published in previous releases.
As a result, latest https://www.npmjs.com/package/ember-cli-fastboot?activeTab=code points to `"fastboot": "4.1.5"` and `"fastboot-express-middleware": "4.1.2"`

whereas https://www.npmjs.com/package/fastboot-express-middleware?activeTab=code points to `"fastboot": "4.1.2"` so we end up using mix of `fastboot` libraries.

@mansona pushed fix for `pnpm-lock.yaml` file to make CI happy.

SergeAstapov:
@mansona this was updated according to discussion in #951

PR #952, opened 2024-10-11
Remove now unused release-it dependencies

SergeAstapov:
@mansona if I'm not mistaken, we just missed to do this in #937


mansona (approved):
you're totally right! I also removed the unused/irrelevant RELEASE.md files on this branch too 👍

PR #955, opened 2025-11-10
Fix ember v6 deprecate-import-onerror-from-ember deprecation

felixbuenemann:
Patches the folllowing deprecation warnings:

* deprecate-import-onerror-from-ember
* importing-inject-from-ember-service
* deprecate-import-view-utils-from-ember

### DIFF
--- a/packages/ember-cli-fastboot/addon/locations/none.js
+++ b/packages/ember-cli-fastboot/addon/locations/none.js
@@ -1,7 +1,7 @@
 /* eslint-disable ember/no-classic-classes, ember/no-get, ember/require-computed-property-dependencies, prettier/prettier */
 import { computed, get } from '@ember/object';
 import { bool, readOnly } from '@ember/object/computed';
-import { inject as service } from '@ember/service';
+import { service } from '@ember/service';
 import { getOwner } from '@ember/application'
 import NoneLocation from '@ember/routing/none-location'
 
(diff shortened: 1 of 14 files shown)


felixbuenemann:
I don't know how to write a version of the code that supports both `import { service } from '@ember/service'` and `import { inject as service } from '@ember/service'` at the same time. Maybe one of the maintainers has an answer.

felixbuenemann:
One option would be to only keep the `deprecate-import-view-utils-from-ember` deprecation fix, which fixes the deprecation message in the browser that can't be silenced by `app/deprecation-workflow.js` and rely on that for the other two deprecations for now.

packages/ember-cli-fastboot/addon/locations/none.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable ember/no-classic-classes, ember/no-get, ember/require-computed-property-dependencies, prettier/prettier */
 import { computed, get } from '@ember/object';
 import { bool, readOnly } from '@ember/object/computed';
-import { inject as service } from '@ember/service';
+import { service } from '@ember/service';
```
SergeAstapov: @felixbuenemann for https://github.com/ember-fastboot/ember-cli-fastboot/pull/955#issuecomment-3511131216 I think it goes like this

```suggestion
import serviceModule from '@ember/service';

const service = serviceModule.service ?? serviceModule.inject;
```

felixbuenemann: Thank you, it must be `import * as serviceModule from '@ember/service'` but then it works.

felixbuenemann:
I added the `import * as serviceModule from '@ember/service'` workaround to support older Ember.js versions, as suggested by @SergeAstapov, but now the packaging tests fail with:

> Using `import * as serviceModule from '@ember/service'` is not supported.

I'm not sure what to do about that other than dropping support for older Ember.js versions.

mansona:
Thanks for the work here, all of this has now been fixed and fastboot is now compatible with Ember 7 🎉 let me know if you have any other issues 👍

PR #956, opened 2026-01-10
Update the default sandbox

NullVoxPopuli:
People run in to this stuff missing all the time.

We should provide this stuff out of the box.

### DIFF
--- a/packages/fastboot/src/sandbox.js
+++ b/packages/fastboot/src/sandbox.js
@@ -21,12 +21,32 @@ module.exports = class Sandbox {
       {
         sourceMapSupport,
         console,
+        fetch,
+        document,
+        navigator,
         setTimeout,
         clearTimeout,
         URL,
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
Lints are failing, but also, the project isn't set up for tooling to support node 16... so... :shrug:

mansona (approved):

mansona:
turns out that 3 of the new globals you added (document, MutationObserver, and ResizeObserver) are not available to us in node (because they just don't make any sense)

I removed Navigator too because that was added in a very very limited capacity in Node 21 and a) we still support Node 20 for now and b) we should discuss if we want to add this because of its limitations in a separate PR 👍

NullVoxPopuli:
more reason fastboot's whole current approach is bad.

It's perfectly reasonable for these things to exist at app boot

PR #967, opened 2026-07-22
add full test matrix to test scenarios

mansona:
With some of the recent changes I wanted to double check that they are still working for the full test suite 👍


mansona:
As this is expanding test coverage (for older Ember versions) I'm going to self-merge 👍

